PR #94, opened 2021-04-26
fix(examples/panorama): input for priv ip

jabielecki:
## Description

Subnet cidr is an input, but can only be modified if priv ip is not
hardcoded. Making it an input as well.

Minor fixes for the module, non-breaking.

### DIFF
--- a/examples/panorama/outputs.tf
+++ b/examples/panorama/outputs.tf
@@ -7,3 +7,8 @@ output "panorama_admin_password" {
   description = "Panorama administrator's initial password."
   value       = random_password.this.result
 }
+
+output "username" {
+  description = "Panorama administrator's initial username."
+  value       = var.username
+}
--- a/modules/panorama/README.md
+++ b/modules/panorama/README.md
@@ -6,47 +6,30 @@ A terraform module for deploying a working Panorama instance in Azure.
 
 ```hcl
 module "panorama" {
-  source = "github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules//modules/panorama"
+  source  = "PaloAltoNetworks/vmseries-modules/azurerm//modules/panorama"
+  version = "0.1.0"
 
   panorama_name       = var.panorama_name
-  name_prefix         = var.name_prefix
-  resource_group_name = var.resource_group_name
-  location            = var.location //Optional; if not provided, will use Resource Group location
-  avzone              = var.avzone   // Optional Availability Zone number
+  resource_group_name = azurerm_resource_group.this.name
+  location            = var.location
+  avzone              = var.avzone // Optional Availability Zone number
 
-  interface = {                          // Only one interface in Panorama VM is supported
-    mgmt = {
-      subnet_id            = ""
-      private_ip_address   = "10.0.0.6" // Optional: If not set, use dynamic allocation
-      public_ip            = "true"    // (optional|bool, default: "false")
-      enable_ip_forwarding = "false"  // (optional|bool, default: "false")
-      primary_interface    = "true"
+  interface = [ // Only one interface in Panorama VM is supported
+    {
+      name               = "mgmt"
+      subnet_id          = var.subnet_id
+      public_ip          = true
+      public_ip_name     = "panorama"
     }
-  }
+  ]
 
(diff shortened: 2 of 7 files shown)


modules/panorama/README.md
```diff
@@ -6,47 +6,30 @@ A terraform module for deploying a working Panorama instance in Azure.
 
 ```hcl
 module "panorama" {
-  source = "github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules//modules/panorama"
+  source  = "PaloAltoNetworks/vmseries-modules/azurerm//modules/panorama"
+  version = "0.1.0"
```
jabielecki, line 10: We can now use the Registry syntax. It would be best to update all the snippets with this one. The previous github syntax has no version, which is horribly unstable.

modules/panorama/README.md
```diff
@@ -6,47 +6,30 @@ A terraform module for deploying a working Panorama instance in Azure.
...
+  resource_group_name = azurerm_resource_group.this.name
+  location            = var.location
+  avzone              = var.avzone // Optional Availability Zone number
+
+  interface = [ // Only one interface in Panorama VM is supported
+    {
+      name               = "mgmt"
+      subnet_id          = var.subnet_id
+      public_ip          = true
+      public_ip_name     = "panorama"
```
jabielecki, line 22: The snippet was outdated by a lot.

modules/panorama/main.tf
```diff
@@ -15,14 +15,14 @@ resource "azurerm_network_interface" "this" {
...
   resource_group_name  = var.resource_group_name
-  enable_ip_forwarding = lookup(var.interface[0], "enable_ip_forwarding", "false")
+  enable_ip_forwarding = lookup(var.interface[0], "enable_ip_forwarding", false)
 
   ip_configuration {
     name                          = var.interface[0].name
     subnet_id                     = var.interface[0].subnet_id
     private_ip_address_allocation = lookup(var.interface[0], "private_ip_address", null) != null ? "static" : "dynamic"
     private_ip_address            = lookup(var.interface[0], "private_ip_address", null) != null ? var.interface[0].private_ip_address : null
-    public_ip_address_id          = lookup(var.interface[0], "public_ip", "false") != "false" ? try(azurerm_public_ip.this[0].id) : null
```
jabielecki, line 25: The `try(1)` is exactly the same as `1`. For the try() to do anything,
it needs to have two or more arguments: `try(1, null)` or `try(1, 2, 3,
4)`

modules/panorama/main.tf
```diff
@@ -15,14 +15,14 @@ resource "azurerm_network_interface" "this" {
   name                 = var.interface[0].name
   location             = var.location
   resource_group_name  = var.resource_group_name
-  enable_ip_forwarding = lookup(var.interface[0], "enable_ip_forwarding", "false")
+  enable_ip_forwarding = lookup(var.interface[0], "enable_ip_forwarding", false)
```
jabielecki, line 18: Avoid `"false"` which has type `string`, to be compared to `false`, which has
type `bool`. This way if user uses `"False"` capitalized the behavior
will be saner.

modules/panorama/README.md
kbechler: 👍

...but ;-) in examples related to a single module only, maybe we could use "latest" tag or sth like this?

examples/panorama/outputs.tf
```diff
@@ -7,3 +7,8 @@ output "panorama_admin_password" {
   description = "Panorama administrator's initial password."
   value       = random_password.this.result
 }
```
kbechler, line 9: Missing "sensitive"

modules/panorama/README.md
jabielecki: Ummm... I'd rather not?

examples/panorama/outputs.tf
jabielecki: I'd rather not mix different changes. This PR is not for 0.15 preparation.

kbechler (approved):

thedevilsvoice (approved):
Please tag with JIRA ticket and mention how you tested this change.

PR #95, opened 2021-04-27
fix(module/vmseries): Destroy issue with vmseries module and loadbalancer backend

lstadnik:
## Description

Due to issue with not proper clean-up resources the terraform raised error.

```
Error: Error waiting for removal of Backend Address Pool Association for NIC "fw00-private" (Resource Group "luca2RG"): Code="OperationNotAllowed" Message="Operation 'startTenantUpdate' is not allowed on VM 'pantffw00' since the VM is marked for deletion. You can only retry the Delete operation (or wait for an ongoing one to complete)." Details=[]
```

## Motivation and Context

It could effect automation lifecycle used by CI/CD tools.

## The alternatives
The quick workaround:
```
terraform destroy -parallelism=1
```

## How Has This Been Tested?

The tests was performed on default examples values from transit-vnet-common.

### DIFF
--- a/modules/vmseries/main.tf
+++ b/modules/vmseries/main.tf
@@ -111,6 +111,8 @@ resource "azurerm_virtual_machine" "this" {
     type         = var.identity_type
     identity_ids = var.identity_ids
   }
+
+  depends_on = [azurerm_network_interface_backend_address_pool_association.this]
 }
 
 resource "azurerm_application_insights" "this" {


modules/vmseries/main.tf
```diff
@@ -111,6 +111,8 @@ resource "azurerm_virtual_machine" "this" {
     type         = var.identity_type
     identity_ids = var.identity_ids
   }
+  
+  depends_on = [ azurerm_network_interface_backend_address_pool_association.this ]
```
jabielecki, line 115: I'm doubtful of any explicit depends_on by now. Honestly it has bitten me many times. I'm worried if it will cause any strange problems down the road or not.

kubanczyk (changes requested):
<apologies, my bad, please disregard this>

jabielecki (changes requested):
It's a cyclic explicit dependency. These have been a problem so many times before, that I need to default to veto. Guilty until proven innocent.

modules/vmseries/main.tf
lucap01: Let's operate on facts, not feelings. Our vmseries with loadbalancer modules doesn't teardown properly. Point and document any use case that it might raise other issue or just write an issue to azurerm repo and ask them for repair, but for now we need fix for it.

jabielecki: All right, you seem optimistic about it after all. What situations did you test? I mean possibilities like: Destroys? Modifications? Additions (i.e. add new interface with lb_ba_pool when some interfaces already exist)? Were there any errors?

When I have not seen any tests, I have defaulted to veto. Don't get me wrong, if this approach actually solves the problem and does not add a worse problem, I will decidedly agree and change my vote.

thedevilsvoice (commented):
This one has been sitting a month. Could you please re-base and re-test if we are planning on merging it? Otherwise let's close it out?

thedevilsvoice:
I found [this in the Terraform documentation](https://www.terraform.io/docs/language/meta-arguments/depends_on.html):
```
The depends_on argument should be used only as a last resort. When using it, always include a comment explaining why it is being used, to help future maintainers understand the purpose of the additional dependency.
```

I think we should avoid this, but my reasoning may be a little different than the rest of the discussion.

migara:
This issue no longer exist

PR #97, opened 2021-04-28
TERRAM-92 ci: add pre-commit framework as a continuous check

jabielecki:
### ci: add pre-commit framework as a continuous check

Adding `pre-commit` to CI makes it verify whether new commits:

- update all the READMEs using `terraform-docs` properly,
- adhere to `tflint` rules.

Therefore it ensures the `develop` branch is clean: future contributors
are less likely to face unexpected pre-commit discrepancies/fails.
The downside is it adds ~20-30 seconds to a CI job.

The current formatting check (`terraform fmt`) is replaced by the
same check done via pre-commit, to more exactly replicate what
a dev does locally.

A specific version of `terraform-docs` is pinned, because recent history
shows they introduce breaking changes from time to time.

### Also, parallelize curl downloads

An enormous speedup of the GitHub Action.
Why is it downloading three alternative Terraform versions:

- latest compatible with all
- oldest compatible with all
- oldest compatible with any


damianfedeczko (approved):

damianfedeczko:
LGTM.

thedevilsvoice (commented):
similar comments to the same PR in the google repo. 
We might consider a single version of this (perhaps in a common repo, or the best practices repo?) so that it is not getting out of sync between repos. 

@migara

PR #100, opened 2021-05-18
feat!: Mark selected outputs as sensitive, allow tf-0.15, require azurerm-2.46+

jabielecki:
### feat: Mark selected outputs as sensitive

Pass the `sensitive` status marked by providers onto the outputs. This
will be a hard requirement from Terraform version 0.15 onwards.

The sensitive outputs are best viewed with `-json` flag.

For docs, remove the `$ ` prompts because of the linting rule:
"MD014/commands-show-output: Dollar signs used before commands without
showing output"

### feat: Allow terraform 0.15.x for all modules and examples

This is the recommended version, so allowing it is a major use case.

### feat(loadbalancer)!: Require azurerm provider version 2.46+

When using provider older than 2.46, it visibly warns:
`"resource_group_name": [DEPRECATED] in resource "azurerm_lb_backend_address_pool"`

Since the examples cannot stay on 2.42 now, pinning them
to the latest version, which is 2.58 (as of now).

It's debatable whether this change is actually Breaking. But for now, I stay on the safe side here.

## How Has This Been Tested?

Apply all the examples.

## Types of changes

## Checklist

- [x] I have marked all the changes that users need to see in the Changelog.
- [x] I have *not* marked the Changelog-unworthy commits as `feat`, `fix`, or `!`.

### DIFF
--- a/examples/bootstrap/README.md
+++ b/examples/bootstrap/README.md
@@ -12,9 +12,16 @@ The following resources will be deployed when using the provided example:
 
 Create a `terraform.tfvars` file and copy the content of `example.tfvars` into it, adjust the variables (in particular the `storage_account_name` should be unique).
 
-```bash
-$ terraform init
-$ terraform apply
+```sh
+terraform init
+terraform apply
+terraform output -json
+```
+
+## Cleanup
+
+```sh
+terraform destroy
 ```
 
 <!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
--- a/examples/bootstrap/README.md
+++ b/examples/bootstrap/README.md
@@ -29,7 +29,7 @@ terraform destroy
 
 | Name | Version |
 |------|---------|
-| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >=0.12.29, <0.15 |
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >=0.12.29, <0.16 |
 | <a name="requirement_azurerm"></a> [azurerm](#requirement\_azurerm) | ~>2.42 |
 | <a name="requirement_random"></a> [random](#requirement\_random) | ~>3.0 |
 
--- a/examples/bootstrap/outputs.tf
+++ b/examples/bootstrap/outputs.tf
@@ -1,24 +1,29 @@
 output "storage_account_name" {
   description = "Name of the Azure Storage Account object used for the Bootstrap."
   value       = module.bootstrap.storage_account.name
+  sensitive   = true
 }
 
 output "storage_account_id" {
   description = "Identifier of the Azure Storage Account object used for the Bootstrap."
   value       = module.bootstrap.storage_account.id
+  sensitive   = true
 }
 
 output "storage_share_name" {
   description = "Name of the File Share within Azure Storage."
   value       = module.bootstrap.storage_share.name
+  sensitive   = true
 }
 
 output "storage_share_id" {
   description = "Identifier of the File Share within Azure Storage."
   value       = module.bootstrap.storage_share.id
+  sensitive   = true
 }
 
 output "primary_access_key" {
   description = "The primary access key for the Azure Storage Account."
   value       = module.bootstrap.primary_access_key
+  sensitive   = true
 }
--- a/examples/bootstrap/versions.tf
+++ b/examples/bootstrap/versions.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">=0.12.29, <0.15"
+  required_version = ">=0.12.29, <0.16"
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
--- a/examples/panorama/versions.tf
+++ b/examples/panorama/versions.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">=0.13, <0.15"
+  required_version = ">=0.13, <0.16"
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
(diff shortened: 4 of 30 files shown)


modules/loadbalancer/main.tf
```diff
@@ -87,9 +87,8 @@ locals {
 }
 
 resource "azurerm_lb_backend_address_pool" "lb_backend" {
-  name                = coalesce(var.backend_name, var.name)
-  resource_group_name = var.resource_group_name
```
jabielecki, line 91: Required before 2.46, produces a warning on 2.46+.

examples/bootstrap/README.md
```diff
@@ -12,17 +12,24 @@ The following resources will be deployed when using the provided example:
 
 Create a `terraform.tfvars` file and copy the content of `example.tfvars` into it, adjust the variables (in particular the `storage_account_name` should be unique).
 
-```bash
-$ terraform init
-$ terraform apply
+```sh
+terraform init
+terraform apply
+terraform output -json
```
jabielecki, line 18: The sensitive outputs are best viewed with `-json` flag.

examples/bootstrap/outputs.tf
```diff
@@ -1,24 +1,29 @@
 output "storage_account_name" {
   description = "Name of the Azure Storage Account object used for the Bootstrap."
   value       = module.bootstrap.storage_account.name
+  sensitive   = true
```
jabielecki, line 4: If we pass an entire object and some attributes are provider-marked as sensitive, the 0.15 requires the entire object to be now sensitive.

It calls for adding a separate single non-sensitive module's output (one object with attributes manually copied) or multiple outputs (each attribute becomes a separate module's output).

examples/bootstrap/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">=0.12.29, <0.15"
+  required_version = ">=0.12.29, <0.16"
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
```
damianfedeczko, line 5: Why aren't we pinning the `azurerm` provider to version `2.58` here? From what I've understood, every version older than 2.46 will be pinned that way. Please ignore the following comments if I didn't understood your intention correctly.

examples/panorama/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">=0.13, <0.15"
+  required_version = ">=0.13, <0.16"
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
```
damianfedeczko, line 5: Ditto.

modules/bootstrap/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">=0.12.29, <0.15"
+  required_version = ">=0.12.29, <0.16"
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
```
damianfedeczko, line 5: Ditto.

modules/panorama/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">=0.12.29, <0.15"
+  required_version = ">=0.12.29, <0.16"
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
```
damianfedeczko, line 5: Ditto.

modules/vmseries/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">=0.12.29, <0.15"
+  required_version = ">=0.12.29, <0.16"
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
```
damianfedeczko, line 5: Ditto.

modules/vmss/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">=0.12.29, <0.15"
+  required_version = ">=0.12.29, <0.16"
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
```
damianfedeczko, line 5: Ditto.

modules/vnet/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">=0.12.29, <0.15"
+  required_version = ">=0.12.29, <0.16"
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
```
damianfedeczko, line 5: Ditto.

damianfedeczko (commented):
Based on the PR description:
"feat(loadbalancer)!: Require azurerm provider version 2.46+" 
I've marked the places where `azurerm` provider has not been updated to the target version.

damianfedeczko:
@jabielecki apologies for the comments style, I couldn't select the line below `source  = "hashicorp/azurerm"` to mark my intent of getting your attention on the `azurerm` provider version... Scusi.

jabielecki:
> @jabielecki apologies for the comments style, I couldn't select the line below `source = "hashicorp/azurerm"` to mark my intent of getting your attention on the `azurerm` provider version... Scusi.

Yeah I understand which line you mean, simply Github wants you to only comment on the lines that are very close to the pull-request under discussion. (Which totally makes sense.)

I don't think we should or could pin the version for these places. These modules are used within bigger setups, so if they are compatible with a range of versions, we allow these versions. Some other piece of the same deployment (say, some Azure module or customer's own module) might require 2.47 for example and we allow that.

damianfedeczko (approved):

PR #104, opened 2021-05-24
feat(examples/transit_vnet_dedicated): new example

jabielecki:
## Description

Converting transit_vnet_common code into transit_vnet_dedicated.
The difference is that dedicated option has separate firewall
pair for inbound traffic.

## How Has This Been Tested?

* terraform init
* terraform apply
* Pass traffic inbound and obew.
* terraform destroy -parallelism 1

## Types of changes

## Checklist


damianfedeczko (approved):
Awesome job!

examples/transit_vnet_dedicated/versions.tf
```diff
@@ -0,0 +1,17 @@
+terraform {
+  required_version = ">=0.13, <0.16"
```
thedevilsvoice, line 2: 13,14,and 15, neat!

examples/transit_vnet_dedicated/variables.tf
```diff
@@ -0,0 +1,211 @@
...
+
+variable "vnet_tags" {
+  description = "A mapping of tags to assign to the created virtual network and other network-related resources. By default equals to `inbound_vmseries_tags`."
+  type        = map(any)
+  default     = {}
+}
+
+variable "olb_private_ip" {
+  description = "The private IP address to assign to the outbound load balancer. This IP **must** fall in the `private_subnet` network."
+  default     = "10.110.0.21"
```
thedevilsvoice, line 145: this default is just so the example will run properly?

thedevilsvoice (commented):
Very nice!

examples/transit_vnet_dedicated/variables.tf
jabielecki: Yes, a static IP works is desired here. Potentially every spoke in its own tfstate is referencing it, so we need to be double sure it never changes (it would be an interruption to production traffic).

PR #110, opened 2021-08-09
feat: allow usage of Terraform versions 1.x

jabielecki:
## Description

After testing 1.0.3 and pre-release 1.1, lets bump the version requirement to allow all 1.x versions of Terraform.

Also autoformat our source using `terraform fmt` v1.0. The `fmt` behavior is different (although compatible) between the old
0.12.31 version and the new 1.0.

For some examples, prefer variable types `list(string)` over `list(any)` and similar.

Improve wording of many input descriptions.

## Motivation and Context

Users tend to expect the 1.0 version of Terraform to be the preferred one, since it's recommended by the vendor.

## How Has This Been Tested?

`apply` of examples

## Types of changes

## Checklist


damianfedeczko (commented):
I see that we have the Azure provider version varies from 2.26 to 2.58 - how do you feel about trying to unify that in a different PR? I can pick it ip.

Besides - 🎖️

jabielecki:
> 
> 
> I see that we have the Azure provider version varies from 2.26 to 2.58 - how do you feel about trying to unify that in a different PR? I can pick it ip.
> 
> Besides that - 🎖️

Thank you a lot for reviewing! I'd suggest using 2.69 based on my recent experiences.

PR #112, opened 2021-08-16
feat: add input `enable_zones` to most modules

jabielecki:
## Description

For azurerm provider >= 2.63 it became impossible to create a Public IP in some
regions, like East Asia. It makes our code unusable in such scenario. To make such regions work, I'm adding a new input
`enable_zones`.

The current default means nothing changes on the usual regions like US East 2.

There is no way to also make possible older providers to support
the same code, so I'm bumping the `version` contraints.

Also fix module panorama:

Before, omitting `zone` on `logging_disks` always resulted in
`Error: expected "zones.0" to not be an empty string` and thus it
had effectively no working default value.

Now the same input will default to the value of input `avzone`,
so that the disk is placed together with the owning virtual machine
and not susceptible to accidental zone separations.

## How Has This Been Tested?

The `terraform apply` and `terraform destroy` of the existing examples as is with the default tfvars.

Then `apply` of the same examples with a change in tfvars:

```
location = "East Asia" # East Asia does not currently support AZ
enable_zones = false
```

## Types of changes

## Checklist


examples/panorama/main.tf
```diff
@@ -80,37 +80,36 @@ module "panorama" {
...
       size : "2048"
-      zone : "1"
       lun : "1"
     }
     disk_name_2 = {
       size : "2048"
-      zone : "2"
       lun : "2"
     }
   }
```
jabielecki, line 110: There is a fix on the module to allow a default `zone` here.

modules/panorama/main.tf
```diff
@@ -94,7 +95,10 @@ resource "azurerm_managed_disk" "this" {
   storage_account_type = "Standard_LRS"
   create_option        = "Empty"
   disk_size_gb         = lookup(each.value, "size", "2048")
-  zones                = [lookup(each.value, "zone", "")]
+  zones = try(
+    [each.value.zone],
+    var.enable_zones && var.avzone != null && var.avzone != "" ? [var.avzone] : null
+  )
```
jabielecki, line 101: Before, omitting zone on logging_disks always resulted in
Error: expected "zones.0" to not be an empty string and thus it
had effectively no working default value.

Now the same input will default to the value of input avzone,
so that the disk is placed together with the owning virtual machine
and not susceptible to accidental zone separations.

jabielecki (commented):
Inline explanations.

damianfedeczko (approved):
Great job 👍

scripts/install.sh
```diff
@@ -10,17 +10,19 @@ set -euo pipefail
 
 cd "$(dirname $0)"
 
-curl -sL https://github.com/terraform-docs/terraform-docs/releases/download/v0.12.1/terraform-docs-v0.12.1-linux-amd64 > terraform-docs    & \
+curl -sL https://github.com/terraform-docs/terraform-docs/releases/download/v0.15.0/terraform-docs-v0.15.0-linux-amd64.tar.gz > terraform-docs.tar.gz    & \
 curl -sL https://github.com/tfsec/tfsec/releases/download/v0.34.0/tfsec-linux-amd64 > tfsec    & \
-curl -sL https://github.com/terraform-linters/tflint/releases/download/v0.20.3/tflint_linux_amd64.zip > tflint.zip    & \
+curl -sL https://github.com/terraform-linters/tflint/releases/download/v0.29.0/tflint_linux_amd64.zip > tflint.zip    & \
```
devsecfranklin, line 15: Should we do a small PR in other repos to update these versions on tflist and terraform docs ?

devsecfranklin (approved):
LGTM

scripts/install.sh
jabielecki: I've registered internal tickets for that, yes.

PR #115, opened 2021-09-22
feat(bootstrap): add the new input `files_md5` to facilitate dynamic files

jabielecki:
## Description

Add the new input `files_md5`. It is optional, to retain the backward
compatibility.

## Motivation and Context

The old problem was that the `files` could not be auto-generated on the
same Apply as this module. The builtin `md5(file(x))`
fails hard if the file does not exist during Validate phase, which is
earlier than Plan or Apply.

It is a hurdle on the way to two user-visible features:

- to read the `authcodes` from secrets out of some vault within the same Terraform Apply
- also to read the `vm-auth-key` from a Panorama in the same Terraform Apply

Now, users are able to generate the `files` on the spot, provided that
they supply the `files_md5` for them. If the `files` entries correspond
to the files that actually exist before the Terraform run, nothing
changes for these and there is no need to supply `files_md5` at all
(that's the old way used everywhere in the examples).

### Test Backward Compatibility

```
cd examples/bootstrap
```

Create a `terraform.tfvars` file and copy the content of `example.tfvars` into it, adjust the variables (in particular the `storage_account_name` should be unique).

```sh
terraform init
terraform apply
terraform output -json
```

See on the Azure storage explorer whether the remote files have expected content.

```sh
terraform destroy
```

### Test The New Functionality

For the above example, replace module call by this:

```
module "bootstrap" {
  source = "../../modules/bootstrap"

  resource_group_name  = azurerm_resource_group.this.name
  location             = azurerm_resource_group.this.location
  storage_account_name = var.storage_account_name
  files                = { (local_file.this.filename) = "config/test.txt" }
  files_md5            = { (local_file.this.filename) = md5(local_file.this.content) }
}

resource "local_file" "this" {
  filename = "test.txt"
  content  = "abcŁąka"
}
```

Repeat the test as indicated above. Passed.

## Types of changes

## Checklist


modules/bootstrap/variables.tf
```diff
@@ -42,6 +42,12 @@ variable "files" {
   type        = map(string)
 }
 
+variable "files_md5" {
+  description = "Optional map of MD5 hashes. Normally the map should be empty, it is intended to be used for the files created in the same Terraform run as this module, which would not be automatically hashed. The keys of the map should be same or all of the keys of `files` input, while the values should be MD5 hashes of file contents. For example `{\"dir/my.txt\" = \"6f7ce3191b50a58cc13e751a8f7ae3fd\"}`"
```
damianfedeczko, line 46:
```suggestion
  description = "Optional map of MD5 hashes. Normally the map should be empty, it is intended to be used for the files created in the same Terraform run as this module, which would not be automatically hashed. The keys of the map should be the same as the keys of the `files` input, while the values should be MD5 hashes generated from file contents. For example `{\"dir/my.txt\" = \"6f7ce3191b50a58cc13e751a8f7ae3fd\"}`"
```

damianfedeczko:
@jabielecki can you please help me to figure out one scenario?

If I would like, for example, to obtain the `authcode` from Vault - what would be the flow to archive that?

modules/bootstrap/variables.tf
jabielecki: Uh, I've meant something else. Let me know how would the text below work for you?

```suggestion
  description = "Optional map of MD5 hashes of file contents. Normally the map could be all empty, because all the files that exist before the `terraform apply` will have their hashes auto-calculated. This input is necessary only for the selected files which are created/modified within the same Terraform run as this module. The keys of the map should be identical with selected keys of the `files` input, while the values should be MD5 hashes of the contents of that file. For example `{\"dir/my.txt\" = \"6f7ce3191b50a58cc13e751a8f7ae3fd\"}`"
```

jabielecki:
> If I would like, for example, to obtain the `authcode` from Vault - what would be the flow to archive that?

I have this sketch, but I didn't want to put it as an example. It is quite a narrow use case.

```hcl2
variable "bootstrap_use_template" {
  description = "If false, use normal files for `template_files`. If true, these files are generated out of a template possibly overwriting the pre-existing content."
}

resource "local_file" "template" {
  for_each = { for k, v in var.template_files : k => v if var.bootstrap_use_template && contains(["license/authcodes"], v) }

  filename          = each.key
  sensitive_content = data.azurerm_key_vault_secret.inbound_fw_authcode.value
}

# Another variant of usage - use the interpolation.
module "template" {
  source = "../../modules/bootstrap"

  resource_group_name  = azurerm_resource_group.this.name
  location             = azurerm_resource_group.this.location
   storage_account_name = var.storage_account_name
  storage_share_name       = var.template_storage_share_name
  files                    = var.template_files
  files_md5                = { for k, v in local_file.template : k => md5(v.sensitive_content) }
  # Because the files are modified within the same Terraform pass, it is necessary to calculate their md5 hash from
  # content. The alternative, calculating from disk, would execute too early and thus have an incorrect hash.
}
```

modules/bootstrap/README.md
```diff
@@ -9,6 +9,33 @@ The module does *not* configure the bootstrap images, licenses, or configuration
...
+│   The given key does not identify an element in this collection value.
+│ - Invalid function argument (at ../../modules/bootstrap/main.tf:68,49-57)
+│   Invalid value for "path" parameter: no file exists at test.txt; this function works only with files that are distributed as part of the configuration source code, so if this file will be created by a resource in this configuration you must instead obtain this result from an attribute of that resource.
+│
+│ At least one expression must produce a successful result.
+╵
+```
+
+If the file does not exist because it is supposed to be generated by the same Terraform run (such as a `local_file` resource), add the hash of its contents to the input `files_md5`.
+
```
jabielecki, line 38: @damianfedeczko  Any idea how to address that ugliness? Or is that explanation sufficient?

modules/bootstrap/variables.tf
damianfedeczko: Indeed, all clear now, thank you 👍

modules/bootstrap/README.md
```diff
@@ -9,6 +9,33 @@ The module does *not* configure the bootstrap images, licenses, or configuration
 
 See the examples/vm-series directory.
 
+## Known Limitations
+
+The error message below means that one or more of the keys of the `files` input map is not a name of an existing file.
+Check the key for misspellings. For current versions of Terraform it is not possible to make this message more readable:
```
damianfedeczko, line 15:
```suggestion
The error message below indicates that one or more of the keys in the `files` input map are not valid - check the file name and the declared file path for possible misspellings or an invalid path reference.
For current versions of Terraform it is not possible to make this message more readable:
```

Nit.

modules/bootstrap/README.md
damianfedeczko, line 38:
```suggestion
If the file does not exist because it is supposed to be generated by the same Terraform run (such as a `local_file` resource), add the hash of its contents to the input `files_md5`.

For example `{\"dir/my.txt\" = \"6f7ce3191b50a58cc13e751a8f7ae3fd\"}`"
```

damianfedeczko: @jabielecki I feel that this explanation is sufficient - but added 2 small suggestions to consider.

damianfedeczko (approved):
Looks good to me, approving in advance - left 2 suggestions to consider. Thank for answering my questions and adding more explanation 🍺

modules/bootstrap/README.md
jabielecki: Thanks, makes more sense now.

PR #118, opened 2021-10-01
community template updates

devsecfranklin:


jabielecki (approved):
Approving with one nitpick, these templates definitely add value here.

PR #119, opened 2021-10-11
fix(vmseries): disallow IP forwarding through the mgmt interface

jabielecki:
## Description

Prevent the first network interface (the management intrface) from
receiving packets with a non-matching destination IP address.

## Motivation and Context

This is per recommendation of the official Reference Architecture. Also,
there are currently no known use cases which would require that.

## How Has This Been Tested?

By `apply` and `destroy -parallelism 1` of examples per their READMEs.

## Types of changes

While this change is technically a breaking one for anyone who uses the first interface for packet
forwarding (for example as a next hop in Azure route table), it's hard to come up with any real
environment or use-case that would be using this today in the field.

## Checklist


ntwrkguru:
This has broken a deployment for a customer. Any breaking changes should bump the minor rev. and trigger a release note.

PR #122, opened 2021-10-13
fix CONTRIBUTING use current versions of tools

jabielecki:
## Description

Use the same versions of tools in CONTRIBUTING as in
`scripts/install.sh`.

## Motivation and Context

Devs will benefit from consistent behavior between their pre-commit and CI's.

## How Has This Been Tested?

Running the instructions locally.

## Types of changes

## Checklist

- [x] I have read the **CONTRIBUTING** document, duh.


devsecfranklin (commented):
wonder if there is a way to automate these version updates/dependencies

Issue #125, opened 2021-10-20
Expose enable_ip_forwarding as a variable


jpshya:
### Is your feature request related to a problem?

For PRA, during the implementation, all the PA firewall VM has by default enable_ip_forwarding set to true. But a recent commit has made enable_ip_forwarding = false for mgmt VMs. This forcing to make changes to already deployed VMs in PRA prod setup.

### Describe the solution you'd like

If enable_ip_forwarding is exposed as a variable, then the wrapper terraform module can set the value based on what is required for the project. 

### Describe alternatives you've considered

No alternatives as of now.

### Additional context

I think it is better to tag a version for PRA usage, so that existing pipelines will refer to the tag and any changes not specific to PRA wouldn't affect the pipelines. Please have a look at setting up a tag for PRA.

jpshya:
Hello Stephen, I am seeing the same behaviour, but our deployments have already completed with "enable_ip_forwarding = true" for management interface. Is there a design change "enable_ip_forwarding = false" for management interface? Can you give some more information?

jpshya:
It is basically trying to update while value to false for management interface (no forced recreation through TF). My query was what was rationally behind the change now and how it will affect the already implemented functionality for PRA project (We are close to completion of the project and wanted to understand if this change would have any impact). I believe the PaloAlto team were working with PRA & Pythian to get these VMs running through TF in first place.

PR #126, opened 2021-10-22
feat(vmseries): Allow ip_forwarding to be a variable

ntwrkguru:
## Description

Previous versions of the module set `enable_ip_forwarding` to `true`. A subsequent change altered this behavior to always set the first interface (management) to `false` and any other interfaces to `true`. This may have unintended consequences for older deployments. This change introduces this argument as a variable, with the default being to set the first to `false` and others to `true`. This is added as an attribute to the `interfaces` mapping.

Fixes: #125

Signed-off-by: Stephen Steiner <ssteiner@paloaltonetworks.com>

## Motivation and Context

The logical change to set management to `false` and all others to `true` may have unintended consequences for older deployments. This allows that behavior to be overridden.

## How Has This Been Tested?

Tested against a sample module along with the pre-commit tests

## Types of changes

## Checklist


devsecfranklin (approved):
Seems like the only issue was naming on the PR title.

LGTM

PR #130, opened 2021-11-15
Adding support for Public IP Prefix in VMSS

hanfil:
## Description
Added public ip prefix variable in variables.tf file and main.tf

## Motivation and Context
Static IP prefix is needed for outbound traffic when you want to allow the traffic through ACL. 

## How Has This Been Tested?
Added the small changes to my code, and then deployed it to azure.
Have not tested it without any given value.

## Types of changes

## Checklist

### DIFF
--- a/modules/vmss/main.tf
+++ b/modules/vmss/main.tf
@@ -83,6 +83,7 @@ resource "azurerm_linux_virtual_machine_scale_set" "this" {
           name                    = "${var.name_prefix}${var.name_fw_mgmt_pip}"
           domain_name_label       = var.mgmt_pip_domain_name_label
           idle_timeout_in_minutes = 4
+          public_ip_prefix_id     = var.public_ip_prefix_id
         }
       }
     }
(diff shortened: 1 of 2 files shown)


modules/vmss/main.tf
```diff
@@ -83,6 +83,7 @@ resource "azurerm_linux_virtual_machine_scale_set" "this" {
           name                    = "${var.name_prefix}${var.name_fw_mgmt_pip}"
           domain_name_label       = var.mgmt_pip_domain_name_label
           idle_timeout_in_minutes = 4
+          public_ip_prefix_id     = var.public_ip_prefix_id
```
jabielecki, line 86: Hi @hanfil and welcome. Could you rename the new variable to something like `mgmt_pip_prefix_id` to clarify that it gets bound to the *mgmt* IP address, and not to the *public* (dataplane) address?

modules/vmss/variables.tf
```diff
@@ -44,6 +44,11 @@ variable "create_public_pip" {
   type    = bool
 }
 
+variable "public_ip_prefix_id" {
```
jabielecki, line 47: Please run `pre-commit run -a` on the repository. I see that the README.md file did not get updated with the new variable of yours.

jabielecki (changes requested):

PR #138, opened 2022-01-05
refactor(vnet)!: moving ngs rules management into nsg resource

FoSix:
## Description

Put NSG management into one resource (instead of two: NSG itself and NSG rule separately).

## Motivation and Context

Although there is no change in the module behaviour, the code is:
* cleaner
* easier to understand (no flattening, nested for loops etc)
* represents the structure of the NSG variable (NSG rules are managed inside the NSG resource using `dynamic` block)

## How Has This Been Tested?

Two tests were done:
1. a new VNET was set up using the new module (using the example code) to prove that the module is working
2. a VNET was set up with the old module and then an update was run against that resource with the new module. This showed that this is a *BRAKING* change, meaning that in such situation two runs of terraform are required:
2.1. the 1st one removes the old rules
2.2 the 2nd adds the rules again, but using the new code. 

## Types of changes

- Breaking change as described in [How Has This Been Tested](#types-of-changes)

## Checklist

### DIFF
--- a/modules/vnet/main.tf
+++ b/modules/vnet/main.tf
@@ -28,14 +28,14 @@ resource "azurerm_subnet" "this" {
   address_prefixes     = each.value.address_prefixes
 }
 
-resource "azurerm_network_security_group" "this" {
-  for_each = var.network_security_groups
+# resource "azurerm_network_security_group" "this" {
+#   for_each = var.network_security_groups
 
-  name                = each.key
-  location            = try(each.value.location, var.location)
-  resource_group_name = var.resource_group_name
-  tags                = var.tags
-}
+#   name                = each.key
+#   location            = try(each.value.location, var.location)
+#   resource_group_name = var.resource_group_name
+#   tags                = var.tags
+# }
 
 locals {
   nsg_rules = flatten([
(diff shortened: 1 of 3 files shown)


FoSix:
Additional tests were done regarding the Azure API behaviour during NSG rules update. In this situation the `dynamic` block does not act like a `count` loop. There is no list created that is maintained inside the state file. Therefore any change in the (for example) order of rules is not treated by terraform (nor Azure) as a change. 

The `azurerm_network_security_group` resource is using this [AzureRM API](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/network-security-groups/create-or-update). The list of rules comes in as a `body` parameter. This means that the comparison of  the existing rules against the requested ones is done by AzureRM instead of terraform and this leads to terraform reporting this change as in-place update instead of a resource replace.

modules/vnet/main.tf
```diff
@@ -28,14 +28,14 @@ resource "azurerm_subnet" "this" {
   address_prefixes     = each.value.address_prefixes
 }
 
-resource "azurerm_network_security_group" "this" {
-  for_each = var.network_security_groups
+# resource "azurerm_network_security_group" "this" {
```
michalbil, line 31: Please removed commented out lines.

modules/vnet/main.tf
```diff
@@ -28,14 +28,14 @@ resource "azurerm_subnet" "this" {
...
-  resource_group_name = var.resource_group_name
-  tags                = var.tags
-}
+#   name                = each.key
+#   location            = try(each.value.location, var.location)
+#   resource_group_name = var.resource_group_name
+#   tags                = var.tags
+# }
 
 locals {
```
michalbil, line 40: This local value is no longer used, please remove it.

modules/vnet/main.tf
FoSix: fixed

modules/vnet/main.tf
FoSix: fixed

devsecfranklin (approved):
Simpler is better.

LGTM

devsecfranklin:
> Additional tests were done regarding the Azure API behaviour during NSG rules update. In this situation the `dynamic` block does not act like a `count` loop. There is no list created that is maintained inside the state file. Therefore any change in the (for example) order of rules is not treated by terraform (nor Azure) as a change.
> 
> The `azurerm_network_security_group` resource is using this [AzureRM API](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/network-security-groups/create-or-update). The list of rules comes in as a `body` parameter. This means that the comparison of the existing rules against the requested ones is done by AzureRM instead of terraform and this leads to terraform reporting this change as in-place update instead of a resource replace.

Thanks for the effort you put in to testing

migara:
> Additional tests were done regarding the Azure API behaviour during NSG rules update. In this situation the `dynamic` block does not act like a `count` loop. There is no list created that is maintained inside the state file. Therefore any change in the (for example) order of rules is not treated by terraform (nor Azure) as a change.
> 
> The `azurerm_network_security_group` resource is using this [AzureRM API](https://docs.microsoft.com/en-us/rest/api/virtualnetwork/network-security-groups/create-or-update). The list of rules comes in as a `body` parameter. This means that the comparison of the existing rules against the requested ones is done by AzureRM instead of terraform and this leads to terraform reporting this change as in-place update instead of a resource replace.

Are you sure? My state indicates those rules are maintained as a list in the Terraform state

```
  "resources": [
    {
      "mode": "managed",
      "type": "azurerm_network_security_group",
      "name": "example",
      "provider": "provider[\"registry.terraform.io/hashicorp/azurerm\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "id": "/subscriptions/xxxx-xxxxx/resourceGroups/example-resources/providers/Microsoft.Network/networkSecurityGroups/acceptanceTestSecurityGroup1",
            "location": "westeurope",
            "name": "acceptanceTestSecurityGroup1",
            "resource_group_name": "example-resources",
            "security_rule": [
              {
                "access": "Allow",
                "description": "",
```

FoSix:
> Are you sure? My state indicates those rules are maintained as a list in the Terraform state

Yes you are right, but that'a list in a resource, not a list of resources. Therefore it does not behave like a loop with `count`.

I did some more testing, it's to much code to paste it in a comment so I leave it [here](https://github.com/FoSix/terraform-azurerm-vmseries-modules/blob/vlan_module_nsg_simplify/VNET-NSG-testing-scenarios.md).

With this change the code is more readable, so it's easier for someone just to go in and contribute.

On the other hand changing a single rule (not the order, but a change in the rule or adding/deleting one) triggers a whole NSG to be updated. By *whole* I mean replacing all rules in a single update. It's not that precise like with managing rules as a separate resources. The questions is how often this code is used to manage existing infrastructure and can we 'sacrifice' precision in favour of readability :) 

Still, that's just an idea.

migara (approved):
LGMT!

FoSix:
moving to draft - found bugs for scenarios where no NSG rules are defined

additionally, when using this module with inbound load balancer, some rules are being created outside of the VNET module. Tests have to be performed to check if an update will not cause deletion of some of the rules.

FoSix:
After a discussion with @migara this PR will be closed. Proposed functionality is a nice way of handling things with azure, but it misses the flexibility the current approach has. It would also not work in brown field deployments, where NSG rules are partially managed by customer.

Issue #107, opened 2021-06-25
Azure AppGW module Support


torstenstern:
### Is your feature request related to a problem?

No ApplicationGateway module available at the moment

### Describe the solution you'd like

the AppGw is needed to support our reference architecture

PR #148, opened 2022-04-01
feat(module/appgw): new module for Application Gateways

FoSix:
## Description

related to #107. An Application Gateway module supporting both `v1` and `v2` versions.

Configuration options were limited to NGFW use cases. Two most important limitations are:
* no support for WAF rules
* single backend per AppGW instance - this defaults to private IPs of Firewalls' public facing network interfaces.

## Motivation and Context

see #107 

## How Has This Been Tested?

Some test examples were created (some referencing the ref arch) but testing was focused rather on possible misconfiguration. The `azurerm_application_gateway` Terraform resource support almost all configuration properties, unfortunately it does not handle validation. So it is possible to mix props that are mutually exclusive or that do not work with the chosen tier (`v1/v2`). So development and testing was focusing mainly on mitigating that risks. 

## Types of changes

## Checklist


FoSix:
For the moment the `pre-commit` check is failing for unknown reason (`terraform fmt` reports some files that are wrongly formatted). This however is not reproducible locally.
ci_yml_—_vmseries_azure__Workspace_

modules/appgw/variables.tf
```diff
@@ -0,0 +1,228 @@
...
+  default     = null
+  type        = string
+}
+
+variable "ssl_policy_cipher_suites" {
+  description = <<-EOF
+  A List of accepted cipher suites. Required only for `ssl_policy_type` set to `Custom`. 
+  
+  Possible values are:
+  * TLS_DHE_DSS_WITH_AES_128_CBC_SHA
```
pimielowski, line 178: I think that maybe it could be better to provide link to documentation with all values: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/application_gateway#cipher_suites

modules/appgw/variables.tf
```diff
@@ -0,0 +1,228 @@
...
+    
+    - `ssl_certificate_path`
+      - example value: "cert/path"
+      - a path to a certificate in `.pfx` format. Required only for `https` listeners
+    - `ssl_certificate_pass`
+      - example value: "cert_password"
+      - a matching password for the certificate specified in `ssl_certificate_path`
+    ```
+  EOF
+  validation {
```
pimielowski, line 121: Wow, nice to see validation in use :)

modules/appgw/outputs.tf
```diff
@@ -0,0 +1,4 @@
+output "public_ip" {
+  description = "A public IP assigned to the Application Gateway."
+  value       = azurerm_public_ip.this.ip_address
+}
```
pimielowski, line 4: New line :)

pimielowski (approved):
Nice work :)

modules/appgw/variables.tf
FoSix: good point, fixed

modules/appgw/variables.tf
FoSix: setting up the config is to complex to just depend on documentation :)

modules/appgw/outputs.tf
FoSix: 👍

FoSix:
# NOTICE

this PR also modifies `github actions` scripts - dependency on a specific `terraform` version has been removed

migara (approved):
Good job 🚀

PR #156, opened 2022-05-24
fixes #146 adding log analytics workspace mode

### LINKED ISSUE #146: VMSeries module supports only Classic Application Insights
For the moment the `vmseries` module support creation of classic AI only. There are regions where only workspace mode for AI is supported. This causes the module to crash during AI provisioning. 

The module should be extended with a possibility to create an AI in a workspace mode - this implies creation of Log Analytics Workspace resource.

jblaszc:
## Description

Adding support for Application Insights workspace-based mode and setting it as the default.

## Motivation and Context

For the moment the vmseries module support creation of classic AI only. There are regions where only workspace mode for AI is supported. This causes the module to crash during AI provisioning.

Azure recommends to use a workspace-based Application Insights mode now. All new Azure regions introduced after Feb 2021 do not support creating classic AI resources. For the older regions, support for classic AI mode will end on Feb 29th 2024.

This fix sets the workspace-based as the dedault mode if Application Insights is required. In case the Classic mode is needed, it can be forced by setting 'application_insights_mode_workspace' variable to 'false'.

[issue146](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/issues/146)

## How Has This Been Tested?

A vmseries example was built with AI mode set to workspace-based and Classic.

## Types of changes

## Checklist


michalbil:
Please also update PR title to match "conventional commits" requirements.

modules/vmseries/main.tf
```diff
@@ -113,13 +113,24 @@ resource "azurerm_virtual_machine" "this" {
     identity_ids = var.identity_ids
   }
 }
+resource "azurerm_log_analytics_workspace" "this" {
+  count = var.metrics_retention_in_days != 0 && var.application_insights_mode_workspace ? 1 : 0
```
FoSix, line 117: This is just an enhancement idea, but I thought that maybe we could change that creation condition based on retention days count to some `true/false` flag? Like we have in different modules. 
Something like:
```
count = var.create_appinsights && var.application_insights_mode_workspace ? 1 : 0
```

with

```
variable "create_appinsights" {
  default = true
}
```

Being dependent only on number of retention days saves us one variable, but IMO it's not that obvious when using the module.

michalbil: Agreed, I also thought about suggesting such an approach.

Additionally, apart from the conditional var, all other AI/log analytics/metrics variables could potentially be wrapped into a single map.

modules/vmseries/main.tf
```diff
@@ -113,13 +113,24 @@ resource "azurerm_virtual_machine" "this" {
...
+
+  name                = coalesce(var.name_log_analytics_workspace, "${var.name}-workspace")
+  location            = var.location
+  resource_group_name = var.resource_group_name # same RG, so no RBAC modification is needed
+  retention_in_days   = var.metrics_retention_in_days
+  sku                 = "PerGB2018"
```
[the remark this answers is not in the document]
jblaszc: A variable \"log_analytics_workspace_sku\" has been added.

modules/vmseries/variables.tf
```diff
@@ -126,6 +126,18 @@ variable "name_application_insights" {
   type        = string
 }
 
+variable "name_log_analytics_workspace" {
```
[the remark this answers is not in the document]
jblaszc: variable name has been changed as suggested.

modules/vmseries/variables.tf
```diff
@@ -126,6 +126,18 @@ variable "name_application_insights" {
...
+  default     = null
+  description = "Name of the Log Analytics workspace to be created. Can be `null`, in which case a default name is auto-generated."
+  type        = string
+}
+
+variable "application_insights_mode_workspace" {
```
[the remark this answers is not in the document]
jblaszc: variable name has been changed as suggested.

modules/vmseries/variables.tf
```diff
@@ -120,9 +120,33 @@ variable "img_version" {
...
+}
+
+variable "log_analytics_workspace_name" {
+  description = "Name of the Log Analytics workspace to be created. Can be `null`, in which case a default name is auto-generated."
+  default     = null
+  type        = string
+}
+
+variable "log_analytics_workspace_sku" {
+  description = "Azure Log Analytics Workspace mode SKU. The default value is set to \"PerGB2018\""
```
FoSix, line 148: I think a good idea would be to point to MS documentation, to a list of possible values.

modules/vmseries/variables.tf
```diff
@@ -145,7 +169,7 @@ variable "identity_ids" {
 }
 
 variable "metrics_retention_in_days" {
-  description = "Specifies the retention period in days. Possible values are 0, 30, 60, 90, 120, 180, 270, 365, 550 or 730. Defaults to 90. A special value 0 disables creation of Application Insights altogether."
+  description = "Specifies the retention period in days. Possible values are 0, 30, 60, 90, 120, 180, 270, 365, 550 or 730. Defaults to 30 days."
```
FoSix, line 172: You say `defaults to 30` but the default is `null`. I know that you probably meant the Azure defaults, but in this case I would suggest leave a comment like `defaults to Azure default value`. Azure default tend to change w/o notice.

Maybe a link do documentation would also help. 

And please use multiline comment in this case (`EOF` style)

modules/vmss/variables.tf
```diff
@@ -399,4 +399,34 @@ variable "diagnostics_storage_uri" {
...
+}
+
+variable "log_analytics_workspace_name" {
+  description = "Name of the Log Analytics workspace to be created. Can be `null`, in which case a default name is auto-generated."
+  default     = null
+  type        = string
+}
+
+variable "log_analytics_workspace_sku" {
+  description = "Azure Log Analytics Workspace mode SKU. The default value is set to \"PerGB2018\""
```
FoSix, line 429: same thing like for the `vmseries` module

modules/vmseries/variables.tf
michalbil: Agreed. Also, no need for the last sentence, since it's already stated in the line below :)

modules/vmseries/variables.tf
```diff
@@ -120,9 +120,33 @@ variable "img_version" {
...
 
-variable "name_application_insights" {
-  default     = null
+variable "create_application_insights" {
+  description = "Enable usage of the Application Insights. The default vailue is set to 'true'."
+  default     = true
+  type        = bool
+}
+
+variable "application_insights_name" {
```
michalbil, line 129: What do you think about wrapping all configuration options for app_insight/log_analytics into one vairable of map type? All those are related to a single, common purpose in the modules.

examples/transit_vnet_dedicated/versions.tf
```diff
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "= 2.64"
+      version = ">= 2.64"
```
pimielowski, line 6: We bump up Azure version to 3.7.0

examples/vmseries/README.md
```diff
@@ -38,9 +38,16 @@ terraform destroy
...
-| <a name="requirement_azurerm"></a> [azurerm](#requirement\_azurerm) | = 2.64 |
+| <a name="requirement_azurerm"></a> [azurerm](#requirement\_azurerm) | = 2.71 |
 | <a name="requirement_random"></a> [random](#requirement\_random) | ~> 3.0 |
 
+## Providers
+
+| Name | Version |
+|------|---------|
+| <a name="provider_azurerm"></a> [azurerm](#provider\_azurerm) | = 2.71 |
+| <a name="provider_random"></a> [random](#provider\_random) | ~> 3.0 |
```
pimielowski, line 49: Everywhere we use version 3.1 :)

examples/vmseries/versions.tf
```diff
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "= 2.64"
+      version = "= 2.71"
```
pimielowski, line 6: We bump up Azure version to 3.7.0

modules/vmseries/versions.tf
```diff
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "~> 2.64"
+      version = "~> 2.71"
```
pimielowski, line 6: We bump up Azure version to 3.7.0

modules/vmss/versions.tf
```diff
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "~> 2.26"
+      version = "~> 2.71"
```
pimielowski, line 6: We bump up Azure version to 3.7.0

examples/vmseries_scaleset/versions.tf
```diff
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "= 2.64"
+      version = "= 2.71"
```
pimielowski, line 6: We bump up Azure version to 3.7.0

jblaszc:
The issue will be fixed as new PR.

PR #166, opened 2022-06-30
feat(module/bootstrap)!: Enhanced user experience of the module usage

FoSix:
## Description

Small refactor of the `bootstrap` module that follows requirements found in real-life.
Things worth mentioning are:
* unify the storage account name variable - use the same one regardless if the storage is being created or only sourced.
* add File Share quota variable - this still default to 50G but now can be customized
* add File Share access tier variable - for the 3.x provider setting this to `null` causes a share to be recreated during each Terraform run. Defaults to `Cool` for cost limitation (it is only used once, during bootstrap) but can be customised.

The bootstrap example was also modified to show how the module can be used to create a single Storage Account with two File Shares for Transit VNET Ref Arch (based on real requirements).

## Motivation and Context

* unify the storage account name variable - having two variables for storage name was:
  * counter-schema when looking at other modules
  * confusing - we have a variable that controls the module behaviour, `existing_storage_account` was duplicating the logic
* add File Share access tier variable - this was causing issues when updating the infrastructure

## How Has This Been Tested?

The `boostrap` example was used to test the module development. All examples using the module were run to verify backwards compatibility. If braking changes were found, examples were adjusted accordingly. 

## Types of changes

## Checklist


modules/bootstrap/variables.tf
```diff
@@ -1,46 +1,46 @@
...
 }
 
 variable "storage_account_name" {
   description = <<-EOF
-  Default name of the storage account, if creating it. Ignored when `existing_storage_account` object is non-null.
+  Name of the Storage Account, either a new or an existing one (depending on the value of `create_storage_account`).
+
   The name you choose must be unique across Azure. The name also must be between 3 and 24 characters in length, and may include only numbers and lowercase letters.
   EOF
   default     = "pantfstorage"
```
migara, line 13: Let's get rid of the default value here

modules/bootstrap/variables.tf
```diff
@@ -52,20 +52,48 @@ variable "files_md5" {
...
+  ```
   EOF
   default     = {}
   type        = map(string)
 }
 
 variable "storage_share_name" {
-  description = "Name of storage share to be created that holds `files` for bootstrapping."
+  description = "Name of storage File Share to be created that holds `files` for bootstrapping."
   default     = "bootstrapshare"
```
migara, line 69: Let's get rid of the default value here

modules/bootstrap/variables.tf
```diff
@@ -52,20 +52,48 @@ variable "files_md5" {
...
 variable "tags" {
-  description = "Azure tags to apply to the created Storage Account. A map, for example `{ team = \"NetAdmin\", costcenter = \"CIO42\" }`"
+  description = <<-EOF
+  A map of Azure tags to apply to the created Storage Account.
+
+  Example: 
+  ```
+  tags = {
+    team       = "NetAdmin"
+    costcenter = "CIO42"
```
migara, line 93: Perhaps match these tags to generic tags we use throughout the rest of the repo?

modules/bootstrap/variables.tf
```diff
@@ -1,46 +1,46 @@
...
-}
-
 variable "create_storage_account" {
-  description = "If true, create a Storage Account and ignore `existing_storage_account`."
+  description = "If `true`, create a Storage Account."
   default     = true
   type        = bool
 }
 
 variable "storage_account_name" {
```
migara, line 7: It would be nice to add a `validation` block here

modules/bootstrap/variables.tf
FoSix: 👍🏻

modules/bootstrap/variables.tf
FoSix: 👍🏻

modules/bootstrap/variables.tf
FoSix: I actually removed the example values. This is what we have in other modules. This way it matches a pattern.

modules/bootstrap/variables.tf
FoSix: 👍🏻 and tested

examples/bootstrap/example.tfvars
```diff
@@ -1,8 +1,15 @@
-resource_group_name  = "example-rg"
-location             = "East US"
-storage_account_name = "examplebootstrap"
+resource_group_name = "example-rg"
+location            = "West US"
+# storage_account_name       = "examplebootstrap"
```
michalbil, line 3: I assume this one should be uncommented?

examples/panorama/variables.tf
```diff
@@ -14,7 +14,11 @@ variable "storage_account_name" {
   Default name of the storage account to create.
   The name you choose must be unique across Azure. The name also must be between 3 and 24 characters in length, and may include only numbers and lowercase letters.
   EOF
-  default     = "pantfstorage"
+  type        = string
+}
+
+variable "storage_share_name" {
+  description = "Name of storage File Share to be created that holds `files` for bootstrapping."
```
michalbil, line 21: In this case it's not actually true - ie. that it holds files for bootstrap.

examples/panorama/example.tfvars
```diff
@@ -3,6 +3,7 @@ tags                 = { environment = "dev" }
 resource_group_name  = "example-rg"
 vnet_name            = "example-vnet"
 storage_account_name = "examplestorage"
+storage_share_name   = "bootstrapshare"
```
michalbil, line 6:
```suggestion
storage_share_name   = "bootdiagshare"
```

examples/bootstrap/example.tfvars
FoSix: 👍🏻

examples/panorama/variables.tf
FoSix: 👍🏻

kbechler (approved):

examples/transit_vnet_dedicated/main.tf
```diff
@@ -112,12 +112,13 @@ module "bootstrap" {
...
-  existing_storage_account = module.bootstrap.storage_account.name
-  storage_share_name       = var.outbound_storage_share_name
-  files                    = var.outbound_files
+  resource_group_name    = azurerm_resource_group.this.name
+  create_storage_account = false
+  storage_account_name   = module.bootstrap.storage_account.name
+  storage_share_name     = var.outbound_storage_share_name
+  files                  = var.outbound_files
+
+  depends_on = [module.bootstrap]
```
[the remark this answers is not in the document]
FoSix: due to the way terraform resolves dependencies. Even though you use this:
```
existing_storage_account = module.bootstrap.storage_account.name
```
Terraform tries to parallelise execution of both modules, most probably  because the `output` is set to a `local` instead of a resource (regardless the fact that this local refers to a resource under some condition).

So there are situation, where the module that creates only a File Share runs before the actual storage is available. Explicit `depends_on` solves the problem.

examples/bootstrap/main.tf
```diff
@@ -3,11 +3,24 @@ resource "azurerm_resource_group" "this" {
...
+module "obew_bootstrap" {
+  source = "../../modules/bootstrap"
+
+  create_storage_account = false
+  storage_account_name   = module.inbound_bootstrap.storage_account.name
+  storage_share_name     = var.obew_storage_share_name
+  resource_group_name    = azurerm_resource_group.this.name
+  files                  = var.obew_files
+
+  depends_on = [module.inbound_bootstrap]
```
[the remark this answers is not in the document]
FoSix: same as above

modules/bootstrap/variables.tf
```diff
@@ -52,20 +55,37 @@ variable "files_md5" {
...
   default     = {}
   type        = map(string)
 }
 
 variable "storage_share_name" {
-  description = "Name of storage share to be created that holds `files` for bootstrapping."
-  default     = "bootstrapshare"
+  description = "Name of storage File Share to be created that holds `files` for bootstrapping."
+  type        = string
+}
```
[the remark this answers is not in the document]
FoSix: yeah, good one, i'll look for Azure limitations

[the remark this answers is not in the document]
FoSix: [validation added](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/commit/971b922fcd25691216a7590dc6d0577d0046890e)

examples/bootstrap/outputs.tf
```diff
@@ -1,29 +1,31 @@
...
+  value       = module.inbound_bootstrap.primary_access_key
   sensitive   = true
 }
 
-output "storage_share_id" {
-  description = "Identifier of the File Share within Azure Storage."
-  value       = module.bootstrap.storage_share.id
-  sensitive   = true
+output "storage_share_names" {
+  description = "Name of the File Share within Azure Storage."
```
migara, line 18:
```suggestion
  description = "Names of the File Shares within Azure Storage."
```

examples/bootstrap/outputs.tf
```diff
@@ -1,29 +1,31 @@
...
+    obew    = module.obew_bootstrap.storage_share.name
+  }
 }
 
-output "primary_access_key" {
-  description = "The primary access key for the Azure Storage Account."
-  value       = module.bootstrap.primary_access_key
-  sensitive   = true
+output "storage_share_ids" {
+  description = "Identifier of the File Share within Azure Storage."
```
migara, line 26:
```suggestion
  description = "Identifiers of the File Shares within Azure Storage."
```

migara (approved):
LGMT 👏🏽

michalbil (approved):
LGTM! with one last suggestion :)

examples/panorama/example.tfvars
```diff
@@ -1,9 +1,9 @@
 location             = "East US"
 tags                 = { environment = "dev" }
-panorama_name        = "example-panorama"
```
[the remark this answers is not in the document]
FoSix: not sure where this came from  - added back

migara (approved):
Great work!

PR #171, opened 2022-08-02
feat(module/loadbalancer): load balancer outbound rules support

FoSix:
## Description

Add support for outbound load balancing rules.

## Motivation and Context

Based on customers' requests. The use case is when customer would like to push all outgoing traffic through LB's public IP.

## How Has This Been Tested?

The code is based on code used in production. 

## Types of changes

## Checklist


modules/loadbalancer/main.tf
```diff
@@ -1,60 +1,33 @@
...
       create_public_ip              = try(v.create_public_ip, false)
       zones                         = try(v.zones, null)
       subnet_id                     = try(v.subnet_id, null)
       private_ip_address_allocation = try(v.private_ip_address_allocation, null)
       private_ip_address            = try(v.private_ip_address, null)
-      rules                         = try(v.rules, {})
     }
+    if(
+      try(v.create_public_ip, false)
+      || (!try(v.create_public_ip, false) && !can(merge(var.frontend_ips, var.outbound_rules)[v.public_ip_name]))
```
pimielowski, line 29: I think you can declare local variable and reuse that part **merge(var.frontend_ips, var.outbound_rules)**.
In if statement in OR ( || ) part is always dependent on the second argument:


```
> (!try(true,false) && !(true))
false
> (!try(false,false) && !(true))
false
> (!try(false,false) && !(false))
true
```
So you can simplify to ``!can(merge(var.frontend_ips, var.outbound_rules)[v.public_ip_name]))``

IMO If you add new local value it will looks cleaner

pimielowski (commented):
Looks good :)

modules/loadbalancer/main.tf
FoSix: your're right, simplified the bool logic + a local with combined rules was added

pimielowski (approved):

migara (approved):
LGTM 👍🏽

PR #175, opened 2022-08-05
fix!: add app insights worker mode (#146)

### LINKED ISSUE #146: VMSeries module supports only Classic Application Insights

### LINKED ISSUE #174: module/vmss: a bug when deleting an infrastructure that uses metrics for autoscalling
Since upgrade to provider 3.x (although that does not have to be the root cause) when destroying an infra with VMSS + autoscaling a resource is being left behind: `microsoft.alertsmanagement/smartdetectoralertrules`. This resource is not present in the state and it prevent resource group deletion.

A workaround is to set the following provider configuration:

```hcl
provider "azurerm" {
  features {
    resource_group {
      # the setting below is required due to an orphaned AI microsoft.alertsmanagement/smartdetectoralertrules object
      prevent_deletion_if_contains_resources = false
    }
  }
}
```

jblaszc:
## Description

Adding support for Application Insights workspace-based mode and setting it as the default.

BREAKING CHANGE: default behaviour as well as variables used to define Application Insights of the `vmseries` and `vmss` module will change. 

## Motivation and Context

For the moment the vmseries module support creation of classic AI only. There are regions where only workspace mode for AI is supported. This causes the module to crash during AI provisioning.

Azure recommends to use a workspace-based Application Insights mode now. All new Azure regions introduced after Feb 2021 do not support creating classic AI resources. For the older regions, support for classic AI mode will end on Feb 29th 2024.

This fix sets the workspace-based as the default mode if Application Insights is required. In case the Classic mode is needed, it can be forced by setting `app_insights_settings.workspace_mode` variable to `false`.

This PR:
 * closes #146 
 * closes #174 by providing a workaround in Application Settings configuration. 

## How Has This Been Tested?

A vmseries (standalone and scale set) examples were built with AI mode set to workspace-based and Classic.

## Types of changes

## Checklist


modules/vmseries/main.tf
```diff
@@ -106,13 +106,30 @@ resource "azurerm_virtual_machine" "this" {
   }
 }
 
-resource "azurerm_application_insights" "this" {
-  count = var.metrics_retention_in_days != 0 ? 1 : 0
+resource "azurerm_log_analytics_workspace" "this" {
+  count = try(var.app_insights_settings.create, false) && try(var.app_insights_settings.workspace_mode, true) ? 1 : 0
```
pimielowski, line 110: Is `try(var.app_insights_settings.workspace_mode, true)` should be `try(var.app_insights_settings.workspace_mode, false)` ?

modules/vmseries/README.md
```diff
@@ -80,6 +81,7 @@ No modules.
...
...ights instance. Can be `null`, in which case a default name is auto-generated.<br>- `workspace_mode`: (optional\|bool) If `true` the Application Insights mode is set to \"Workspace-based\". Otherwise, the legacy \"Classic\" mode is used.<br>- `metrics_retention_in_days`: (optional\|number) Specifies the retention period in days. Possible values are 0, 30, 60, 90, 120, 180, 270, 365, 550 or 730. Defaults to 90.<br>- `application_type`: (optional\|string) Specifies the type of Application Insights to create. Valid value is `other`.<br>- `log_analytics_name`: (optional\|string) The name of the Log Analytics workspace. Can be `null`, in which case a default name is auto-generated.<br>- `log_analytics_sku`: (optional\|string) Azure Log Analytics Workspace mode SKU. The default value is set to \"PerGB2018\".<br><br>Example:<pre>{<br>      create                    = true<br>      name                      = "AppInsights"<br>      workspace_mode            = true<br>      metrics_retention_in_days = 30<br>      application_type          = "other"<br>      log_analytics_name        = "LogAnalyticsName"<br>      log_analytics_sku         = "PerGB2018"<br>  }</pre> | `map(any)` | `{}` | no |
```
pimielowski, line 84: `metrics_retention_in_days` default is set to null, not 90 
`  retention_in_days   = try(var.app_insights_settings.metrics_retention_in_days, null)`

modules/vmseries/variables.tf
```diff
@@ -188,4 +176,33 @@ variable "avzones" {
...
   type        = list(string)
+}
+
+variable "app_insights_settings" {
+  description = <<-EOF
+  Application Insights variable list:
+  - `create`: (optional|bool) Enable usage of the Application Insights. The default vailue is set to `false`.
+  - `name`: (optional|string) The name of the Applications Insights instance. Can be `null`, in which case a default name is auto-generated.
+  - `workspace_mode`: (optional|bool) If `true` the Application Insights mode is set to \"Workspace-based\". Otherwise, the legacy \"Classic\" mode is used.
+  - `metrics_retention_in_days`: (optional|number) Specifies the retention period in days. Possible values are 0, 30, 60, 90, 120, 180, 270, 365, 550 or 730. Defaults to 90.
```
pimielowski, line 187: ditto

modules/vmss/main.tf
```diff
@@ -158,17 +158,34 @@ resource "azurerm_linux_virtual_machine_scale_set" "this" {
   }
 }
 
-resource "azurerm_application_insights" "this" {
-  count = var.metrics_retention_in_days != 0 ? 1 : 0
+resource "azurerm_log_analytics_workspace" "this" {
+  count = (length(var.autoscale_metrics) != 0 || try(var.app_insights_settings.create, false)) && try(var.app_insights_settings.workspace_mode, true) ? 1 : 0
```
pimielowski, line 162: Is try(var.app_insights_settings.workspace_mode, true) should be try(var.app_insights_settings.workspace_mode, false) ?

pimielowski (commented):
Looks good :)

modules/vmseries/main.tf
jblaszc: The old (standard) app-insights will be depricated on Feb 29th 2024. So the module is using the app-insights with workspaces as the default set. This is the reason of having `True` here.

modules/vmseries/README.md
jblaszc: True, the 90 is the azure default value if you set variable to `null`. But, I will change this description.

modules/vmss/main.tf
jblaszc: Same as the previous comment, the workspace is used by default.

modules/vmseries/variables.tf
```diff
@@ -188,4 +176,33 @@ variable "avzones" {
...
+  Application Insights variable list:
+  - `create`: (optional|bool) Enable usage of the Application Insights. The default vailue is set to `false`.
+  - `name`: (optional|string) The name of the Applications Insights instance. Can be `null`, in which case a default name is auto-generated.
+  - `workspace_mode`: (optional|bool) If `true` the Application Insights mode is set to \"Workspace-based\". Otherwise, the legacy \"Classic\" mode is used.
+  - `metrics_retention_in_days`: (optional|number) Specifies the retention period in days. Possible values are 0, 30, 60, 90, 120, 180, 270, 365, 550 or 730. If not set, Azure defaults it to 90.
+  - `application_type`: (optional|string) Specifies the type of Application Insights to create. Valid value is `other`.
+  - `log_analytics_name`: (optional|string) The name of the Log Analytics workspace. Can be `null`, in which case a default name is auto-generated.
+  - `log_analytics_sku`: (optional|string) Azure Log Analytics Workspace mode SKU. The default value is set to \"PerGB2018\".
+
+  Example:
```
FoSix, line 192: maybe another example with minimum setting would be helpful

FoSix (commented):
👍🏻

modules/vmseries/main.tf
```diff
@@ -106,13 +106,30 @@ resource "azurerm_virtual_machine" "this" {
...
+  name                = try(var.app_insights_settings.name, "${var.name}-AppInsights")
+  location            = var.location
+  resource_group_name = var.resource_group_name # same RG, so no RBAC modification is needed
+  workspace_id        = try(var.app_insights_settings.workspace_mode, true) ? azurerm_log_analytics_workspace.this[0].id : null
+  application_type    = try(var.app_insights_settings.application_type, "other")
+  retention_in_days   = try(var.app_insights_settings.metrics_retention_in_days, null)
+
+  tags = var.tags
+
+  depends_on = [
```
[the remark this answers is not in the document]
jblaszc: yeah, this explicit dependency is not needed here, will be removed.

modules/vmseries/variables.tf
```diff
@@ -188,4 +176,33 @@ variable "avzones" {
   EOF
   default     = []
   type        = list(string)
+}
+
+variable "app_insights_settings" {
+  description = <<-EOF
+  Application Insights variable list:
+  - `create`: (optional|bool) Enable usage of the Application Insights. The default vailue is set to `false`.
```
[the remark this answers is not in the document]
jblaszc: Sounds logical, so the simplest `app_insights_settings` declaration would be `{}`, and this will enable app insights with the default parameters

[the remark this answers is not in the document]
FoSix: yes and no, how to declare that you do not want any AI? `app_insights_settings = null`? or maybe leave the name as the required value?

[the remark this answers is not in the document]
jblaszc: My plan is to set default value to `null`:

```
variable "app_insights_settings" {
  default     = null
  type        = map(any)
}
```
And in the module, the check would look like this:
```
resource "azurerm_log_analytics_workspace" "this" {
  count = var.app_insights_settings != null
```
Now, the part I don't like. To enable the App Insights with the default settings, you need to define empty variable, like:
``` app_insights_settings = {} ```

is this approach OK?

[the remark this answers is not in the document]
Comment (FoSix) [reply modules/vmseries/variables.tf:184]: I would say yes? As long as it is documented. 
On the other hand it is couter-schema, meaning that in other places we do use an empty map **not** to trigger a resource creation.

File: modules/vmseries/variables.tf [hunk 8]
```diff
@@ -208,4 +196,51 @@ variable "diagnostics_storage_uri" {
...
+  
+  If the variable is:
+  - not defined - Application Insights will not be created (default behavior)
+  - defined as empty map `{}` - Application Insights will be created based on the default parameters
+  - defined as a map - Application Insights will be created with the defined properties, for any skipped - a default value will be used.
+
+  Available properties are:
+  - `name`                      - (optional|string) Name of the Applications Insights instance. Can be `null`, in which case a default name is auto-generated.
+  - `workspace_mode`            - (optional|bool)   Application Insights mode. If `true` (default), the "Workspace-based" mode is used. With `false`, the mode is set to legacy "Classic".
+  - `metrics_retention_in_days` - (optional|number) Specifies the retention period in days. Possible values are 0, 30, 60, 90, 120, 180, 270, 365, 550 or 730. Azure defaults is 90.
```
pimielowski, line 213: This value doesn't have default value, either in resource itself since it is a optional value and in module codebase :)

pimielowski (approved):
Looks good 👍  :)

michalbil (approved):

FoSix (approved):
👍🏻

PR #176, opened 2022-08-05
feat(module/panorama): Add disk type variable for Panorama

pimielowski:
## Motivation and Context

The current Panorama module uses a [hardcoded disk type](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/cc425e85f567ea4a377a2b600257ea322c394b6f/modules/panorama/main.tf#L95) for logging disks

## How Has This Been Tested?

Static code tests

## Types of changes

## Checklist


migara (approved):
Nice work! 👏🏽

modules/panorama/variables.tf
```diff
@@ -64,6 +64,12 @@ variable "enable_plan" {
   type        = bool
 }
 
+variable "panorama_disk_type" {
+  description = "Specifies the type of managed disk to create. Possible values are either Standard_LRS, StandardSSD_LRS, Premium_LRS or UltraSSD_LRS."
```
FoSix, line 68: Maybe add validation? Some of `string` values in Azure are case sensitive (especially since provider v. 3.x), so it would make sense to check if the spelling is correct by verifying if the value is one of  `Standard_LRS`, `StandardSSD_LRS`, `Premium_LRS` or `UltraSSD_LRS`.

modules/panorama/main.tf
```diff
@@ -55,7 +55,7 @@ resource "azurerm_virtual_machine" "panorama" {
     name              = coalesce(var.os_disk_name, "${var.panorama_name}-disk")
     caching           = "ReadWrite"
     create_option     = "FromImage"
-    managed_disk_type = "Standard_LRS"
+    managed_disk_type = var.panorama_disk_type
```
FoSix, line 58: shouldn't we add the same functionality to managed disks? For the moment the disk type is hardcoded [there](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/develop/modules/panorama/main.tf#L96).

FoSix (commented):
👍🏻

modules/panorama/variables.tf
pimielowski: Good catch, I add that :)

modules/panorama/main.tf
pimielowski: I add Some logic, but there is a catch - since we add disks via map I add logic to specify disk type for each disks, but I drop validation here (It can look too complex in one line).

PR #182, opened 2022-08-24
fix: add relaxed azure provider version requirements

FoSix:
## Description

The current version requirements for Azure were `~>3.7.0`. Even though `~>` was used, the version has patch value included which prevents using more recent 3.x provider versions.

Also the version requirements were removed from examples, this way we relay on minimum versions from modules. The example is only a glue-code and does not relay on the provider directly. 

## Motivation and Context

We're pinned to Azure provider version 3.7.0.

## How Has This Been Tested?

`terraform init` was run to check if a newer version then 3.7.0 will be used. 

## Types of changes

## Checklist

### DIFF
--- a/examples/bootstrap/README.md
+++ b/examples/bootstrap/README.md
@@ -32,7 +32,6 @@ terraform destroy
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.15, < 2.0 |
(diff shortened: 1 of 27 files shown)


pimielowski (approved):
You are right about removing provider version from glue-code, I think this approach can be applied to other projects.

migara (approved):
LGTM 👍🏽

Issue #187, opened 2022-08-29
examples refactor


FoSix:
Refactor of all examples - refresh the code to match all changes made to modules in recent releases. 
Unify code in all examples.

example | issue | done
--- | --- | --- 
bootstrap | | <li>[ ] done</li>
panorama | | <li>[ ] </li>
transit_vnet_common | #228 | <li>[ ] </li>
transit_vnet_dedicated | #242  | <li>[ ] </li>
vmseries | | <li>[ ] </li>
vmseries_scaleset | | <li>[ ] </li>
vnet | | <li>[ ] </li>

PR #191, opened 2022-09-08
feat(module/vmseries): add ssh key-pair auth

jblaszc:
## Description
Allows for SSH key-pair authentication for the admin user. The code introduces new variable `ssh_key` which should contains a content of an SSH public key. `password` always takes precedence over SSH keys.

## Motivation and Context
Current code supports only username/password authentication and it may be beneficial to have an option to authenticate based on SSH keys. This method is supported for other vendors (like AWS). 

## How Has This Been Tested?
Tested successfully on the `vmseries` example, where instance has been initialized in the following configurations:
- only `password` authentication
- only `SSH key` authentication
- `password` and 'SSH key` authentication

## Types of changes

## Checklist


modules/vmseries/variables.tf
```diff
@@ -90,6 +90,19 @@ variable "password" {
   type        = string
 }
 
+variable "ssh_key_path" {
```
michalbil, line 93: How about not adding a variable for a single key, but rather a set of key data - having only an option for a single key is rather a development/demo use case, in real world there will be multiple users. That's actually how it's done in GCP VM-Series module.

modules/vmseries/main.tf
```diff
@@ -86,7 +86,14 @@ resource "azurerm_virtual_machine" "this" {
   }
 
   os_profile_linux_config {
-    disable_password_authentication = false
+    disable_password_authentication = var.password == null ? true : false
```
michalbil, line 89: I think that `password` var should no longer be a required one (ie. have a default set to `null`). Additionally, it would be good to document the fact that if password auth is disabled, then at least one key has to be specified (https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/virtual_machine#ssh_keys).

jblaszc: good point, I've set `null` as the default for the `password` variable.

modules/vmseries/variables.tf
FoSix: For the moment the code is supporting only a single account. Changing it to setup multiple accounts - that would be a conceptual change.

modules/vmseries/main.tf
FoSix: implemented

modules/vmseries/main.tf
```diff
@@ -86,7 +86,14 @@ resource "azurerm_virtual_machine" "this" {
...
+    dynamic "ssh_keys" {
+      for_each = var.ssh_key != null ? ["one"] : []
```
pimielowski, line 91: We need to think about standardise how we use this statement, sometimes is:
`for_each = var.ssh_key != null ? ["one"] : []` and sometimes it is `for_each = var.ssh_key != null ? [1] : []` :)

pimielowski (approved):
Look's good

modules/vmseries/main.tf
FoSix: This is an old example, `["one"]` is used in most of the loops, hence it follows this standard. But this will be addressed in #187

modules/panorama/main.tf
```diff
@@ -70,7 +70,14 @@ resource "azurerm_virtual_machine" "panorama" {
   }
 
   os_profile_linux_config {
-    disable_password_authentication = false
+    disable_password_authentication = var.password == null ? true : false
```
michalbil, line 73: Same point as in this thread https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/pull/191/files#r984585354

FoSix: fixed

modules/panorama/variables.tf
```diff
@@ -54,10 +54,30 @@ variable "username" {
 }
 
 variable "password" {
-  description = "Initial administrative password to use for Panorama. Mind the [Azure-imposed restrictions](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/faq#what-are-the-password-requirements-when-creating-a-vm)."
+  description = "Initial administrative password to use for Panorama. If not defined the `ssh_key` variable must be specified. Mind the [Azure-imposed restrictions](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/faq#what-are-the-password-requirements-when-creating-a-vm)."
+  default     = null
   type        = string
 }
 
+variable "ssh_key" {
```
michalbil, line 62:
```suggestion
variable "ssh_keys" {
```

modules/vmseries/variables.tf
```diff
@@ -86,10 +86,30 @@ variable "username" {
 }
 
 variable "password" {
-  description = "Initial administrative password to use for VM-Series. Mind the [Azure-imposed restrictions](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/faq#what-are-the-password-requirements-when-creating-a-vm)."
+  description = "Initial administrative password to use for VM-Series. If not defined the `ssh_key` variable must be specified. Mind the [Azure-imposed restrictions](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/faq#what-are-the-password-requirements-when-creating-a-vm)."
+  default     = null
   type        = string
 }
 
+variable "ssh_key" {
```
michalbil, line 94:
```suggestion
variable "ssh_keys" {
```

michalbil (approved):

FoSix (approved):

PR #209, opened 2022-11-16
fix(module/vmseries,module/vmss): removed validation for user data bootstrapping

FoSix:
## Description

Removed validation block for user data bootstrap options in both the vmseries and vmss modules.

## Motivation and Context

The bootstrap options are not fully documented. Having this functionality limits flexibility on the user data bootstrapping usage. And it does not guarantee that the firewall will be correctly configured - the property name can be OK, but it's content not. 

## Types of changes

## Checklist


migara (approved):
💯

PR #210, opened 2022-11-16
ci: release CI/CD

FoSix:
## Description

Introducing release CI automation.

## Motivation and Context

This PR brings in:
* `Makefile`'s for each module and example that introduce a wrapper layer and a common interface to test the code
* PR title lining that will force proper commit names on `main` branch that can be used to calculate release version using semantic versioning
* PR validation workflow that for the moment does TF code validation, it will handle test deployment and end to end tests in the future
* Release workflow responsible for automated releases

## How Has This Been Tested?

All actions were tested on a 1:1 copy of this repository.

## Types of changes

## Checklist


examples/bootstrap/Makefile
```diff
@@ -0,0 +1,29 @@
...
+apply_file:
+	@../../makefile.sh apply_file
+
+apply: init plan_file apply_file
+
+indepotency:
+	@../../makefile.sh indepotency '${TF_PARAMS}'
+
+destroy:
+	@../../makefile.sh destroy '${RG}'
```
pimielowski, line 29: I was thinking of how to deduplicate this part since we have it in all Makefiles.
Maybe we can create one Makefile with all that options and we can pass it through 
`make -f ../Makefile TF_PARAMS='${TF_PARAMS}'` do you thing this is achievable?

pimielowski (approved):
Great works!

examples/bootstrap/Makefile
FoSix: achievable, but not for the moment. Having this duplicated gives you flexibility when moving module by module from plain commands to `terratest`

PR #213, opened 2022-11-29
feat: popping minimum supported TF version to 1.0.0

### LINKED ISSUE #214: Bug(ci): Version 3.9.4 with arch x64 not found
## Describe the bug

python 3.9.4 is no longer supported in GitHub Actions

## Current behavior

Following error is thrown:
```
Version 3.9.4 with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

FoSix:
## Description

Setting minimum supported Terraform version to 1.0.0. Also closes #214.

## Motivation and Context

Main motivation is the fact that Terraform versions lower than 1.x tend to cause errors with Azure TF provider 3.x

## How Has This Been Tested?

No additional tests required as the code is tested with different TF versions during development.

## Types of changes

## Checklist


pimielowski (approved):
Since we are popping into 1.0 in Azure maybe there is a good idea to move to that version in other vmseries modules?

PR #219, opened 2022-12-05
feat(module/vnet): add possibility to re-use existing subnets

FoSix:
## Description

Extend brownfield deployments by re-using not only existing VNET but also existing Subnets.

## Motivation and Context

Based on real life examples where deployment is done in an already existing VNET + Subnets. This module allowed only to re-use an existing VNET and forced creation of Subnets. With this change only NSGs and UDRs are being created and assigned to already existing Subnets. Subnet definition structure does not  change. 

## How Has This Been Tested?

A modified VNET example was used to test the code.

## Types of changes

## Checklist


modules/vnet/main.tf
```diff
@@ -20,14 +20,26 @@ locals {
...
+  for_each = var.create_subnets ? var.subnets : {}
 
   name                 = each.key
   resource_group_name  = var.resource_group_name
   virtual_network_name = local.virtual_network.name
   address_prefixes     = each.value.address_prefixes
 }
 
+data "azurerm_subnet" "this" {
+  for_each = var.create_subnets == false ? var.subnets : {}
```
FoSix, line 32: for whatever reason this throws an error when used in glue code (not observed during tests though):
```
Error: Inconsistent conditional result types
│
│   on .terraform/modules/vnet/modules/vnet/main.tf line 23, in resource "azurerm_subnet" "this":
│   23:   for_each = var.create_subnets ? var.subnets : {}
│     ├────────────────
│     │ var.create_subnets is true
│     │ var.subnets is object with 3 attributes
│
│ The true and false result expressions must have consistent types. The 'true' value includes object attribute "management", which is absent in the 'false' value.
```

migara (approved):

PR #221, opened 2022-12-20
feat(module/natgw): introduce NAT GW module

FoSix:
## Description

This PR introduces a new module that can be used to create a NAT Gateway.

## Motivation and Context

NAT GW is a the only Azure method of handling the outgoing traffic that can prevent SNAT port exhaustion.  Therefore a module that automates creation of this and related resources is introduced with this PR.

This module is capable of:
* creating a new NAT GW with either a Public IP or Public IP prefix 
* creating a new NAT GW and reusing existing Public IP or Public IP Prefix
* only sourcing an existing NAT GW.

In all 3 cases the NAT GW will be associated with subnets provided as a map of subnet IDs.

## How Has This Been Tested?

Sample code based on Common Ref Architecture was deployed. 

## Types of changes

## Checklist

### DIFF
--- a/modules/natgw/main.tf
+++ b/modules/natgw/main.tf
@@ -1,0 +1,96 @@
+resource "azurerm_public_ip" "this" {
+  count = (var.create_natgw && var.create_pip) ? 1 : 0
+
+  name                = "${var.name}-pip"
+  resource_group_name = var.resource_group_name
+  location            = var.location
+  allocation_method   = "Static"
+  sku                 = "Standard"
+  zones               = var.zone != null ? [var.zone] : null
+
+  tags = var.tags
+}
+
+data "azurerm_public_ip" "this" {
+  count = (var.create_natgw && !var.create_pip && var.existing_pip_name != null) ? 1 : 0
+
+  name                = var.existing_pip_name
+  resource_group_name = var.existing_pip_resource_group_name == null ? var.resource_group_name : var.existing_pip_resource_group_name
+}
+
+resource "azurerm_public_ip_prefix" "this" {
+  count = (var.create_natgw && var.create_pip_prefix) ? 1 : 0
+
+  name                = "${var.name}-pip-prefix"
+  resource_group_name = var.resource_group_name
+  location            = var.location
+  ip_version          = "IPv4"
+  prefix_length       = var.pip_prefix_length
+  sku                 = "Standard"
+  zones               = var.zone != null ? [var.zone] : null
+
+  tags = var.tags
+}
+
+data "azurerm_public_ip_prefix" "this" {
+  count = (var.create_natgw && !var.create_pip_prefix && var.existing_pip_prefix_name != null) ? 1 : 0
+
+  name                = var.existing_pip_prefix_name
+  resource_group_name = var.existing_pip_prefix_resource_group_name == null ? var.resource_group_name : var.existing_pip_prefix_resource_group_name
+}
+
+resource "azurerm_nat_gateway" "this" {
+  count = var.create_natgw ? 1 : 0
+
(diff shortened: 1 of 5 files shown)


modules/natgw/main.tf
```diff
@@ -0,0 +1,96 @@
...
+  zones               = var.zone != null ? [var.zone] : null
+
+  tags = var.tags
+}
+
+data "azurerm_public_ip_prefix" "this" {
+  count = (var.create_natgw && !var.create_pip_prefix && var.existing_pip_prefix_name != null) ? 1 : 0
+
+  name                = var.existing_pip_prefix_name
+  resource_group_name = var.existing_pip_prefix_resource_group_name == null ? var.resource_group_name : var.existing_pip_prefix_resource_group_name
```
sebastianczech, line 39: Could we make it simpler e.g. ``coalesce(var.existing_pip_prefix_resource_group_name, var.resource_group_name)`` ?

modules/natgw/main.tf
```diff
@@ -0,0 +1,96 @@
...
+  ip_version          = "IPv4"
+  prefix_length       = var.pip_prefix_length
+  sku                 = "Standard"
+  zones               = var.zone != null ? [var.zone] : null
+
+  tags = var.tags
+}
+
+data "azurerm_public_ip_prefix" "this" {
+  count = (var.create_natgw && !var.create_pip_prefix && var.existing_pip_prefix_name != null) ? 1 : 0
```
sebastianczech, line 36: In case, when ``create_pip_prefix`` is false and ``existing_pip_prefix_name`` is null, we just use ``try(data.azurerm_public_ip_prefix.this[0], null)`` as in line 73 ? Maybe we could simplify it by making ``create_pip_prefix`` local which is true only if variable ``existing_pip_prefix_name`` is empty ?

sebastianczech (commented):
LGTM, I've put 2 tiny comments :)

modules/natgw/main.tf
FoSix: yes, why not. There were some problems with `coalesce` though, can't remember details.

modules/natgw/main.tf
FoSix: 1. I do not follow :) you would have to drop a code suggestion
2. having this logic here gives two benefits:
  * you call this data source only when you need to
  * this condition documents itself (although it is hard to read ;))

If I understood correctly you want to simplify the condition and move part of the logic to a `try` statement further in the code. But this would brake both bullet points.

sebastianczech: I'm sorry for unclear description. I was mainly thinking about simplifying and making it easier to understand that condition by use local - then it's ready to catch that we are using data source if we are creating NAT GW and we want to use public IP prefix e.g.

```suggestion
  count = (var.create_natgw && local.use_pip_prefix) ? 1 : 0
```

where local is true on following condition:

```
locals {
  use_pip_prefix = !var.create_pip_prefix && var.existing_pip_prefix_name != null
}
```

modules/natgw/main.tf
sebastianczech: if there are problems with `coalesce`, let's leave it as it's

modules/natgw/main.tf
FoSix: I would stick to the original code. We only split the logic into to places. If someone really wants to understand what is happening, one has to go to `locals` to figure it out.

modules/natgw/main.tf
FoSix: :+1: did coalesce as suggested

sebastianczech (approved):
LGTM 👍

PR #222, opened 2022-12-27
feat(module/appgw): add domain_name_label support for AppGW Public IP

alperenkose:
## Description

With domain_name_label, Azure will generate a fqdn for the Azure Application Gateway Public IP in <domain_name_label>.<region>.cloudapp.azure.com format.

## Motivation and Context

For testing it makes easy to get an fqdn that resolves to Application Gateway public IP. 

## How Has This Been Tested?

Tested on example code with the domain_name_label attribute and removing it.

## Types of changes

## Checklist

### DIFF
--- a/modules/appgw/main.tf
+++ b/modules/appgw/main.tf
@@ -40,7 +40,7 @@ resource "azurerm_public_ip" "this" {
 
   sku               = "Standard"
   allocation_method = "Static"
-  domain_name_label = try(var.domain_name_label, null)
+  domain_name_label = var.domain_name_label
   zones             = var.zones
   tags              = var.tags
 }
--- a/modules/appgw/main.tf
+++ b/modules/appgw/main.tf
@@ -40,6 +40,7 @@ resource "azurerm_public_ip" "this" {
 
   sku               = "Standard"
   allocation_method = "Static"
+  domain_name_label = try(var.domain_name_label, null)
   zones             = var.zones
   tags              = var.tags
 }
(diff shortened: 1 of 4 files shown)


modules/appgw/main.tf
```diff
@@ -40,6 +40,7 @@ resource "azurerm_public_ip" "this" {
 
   sku               = "Standard"
   allocation_method = "Static"
+  domain_name_label = try(var.domain_name_label, null)
```
sebastianczech, line 43: If we have default value for ``domain_name_label``, then can we skip ``try`` function ?

sebastianczech (commented):
Nice improvement, I've added 1 tiny comment about ``try`` :)

pimielowski (changes requested):
Like @sebastianczech said, you can get rid of the try() function.

migara (approved):

pimielowski (approved):

modules/appgw/outputs.tf
```diff
@@ -2,3 +2,8 @@ output "public_ip" {
   description = "A public IP assigned to the Application Gateway."
   value       = azurerm_public_ip.this.ip_address
 }
+
+output "public_domain_name" {
+  description = "Public domain name assigned to the Application Gateway."
+  value       = azurerm_public_ip.this.fqdn
```
FoSix, line 8: in this form, when `var.domain_name_label` is not specified, you get the following input:
```
fqdn = {
  "public_appgw" = tostring(null)
}
```

Which works, but doesn't look nice ;) 

And we never know how this will be treated in future provider versions. So maybe wrap it in some condition like:

```
var.domain_name_label == null ? null : azurerm_public_ip.this.fqdn
```

FoSix: normal behaviour

PR #226, opened 2023-01-13
fix(module/vmseries): remove count when creating PublicIPs and NICs

FoSix:
## Description

Use `for_each` loop to create Public IPs and Network Interfaces instead of count - introduce `interfaces` map in `outputs` where keys are interface names instead of a list. 

## Motivation and Context

This change causes the code to be more predictable:
* we do not use count for resource creation
* PIPs and NICs can be referenced using interface name, hence it's easier and more intuitive
* the interfaces in output is no longer a list, but a map, where key is the NIC name - this is especially important where we need to reference a particular interface - it's easier to do it by name, then by an index number (not only when trying to write/understand the code, but also when working with other modules, especially the Application Gateway)

## How Has This Been Tested?

All examples using VMSeries module were deployed in order to test the code.

## Types of changes

## Checklist


migara (approved):

pimielowski (approved):
Looks good :) 👍

PR #106, opened 2021-05-27
my first awesome commit

ancoleman:
## Types of changes

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes if appropriate.
- [ ] All new and existing tests passed.

PR #229, opened 2023-01-20
fix(example/transit_vnet_common): add code flexibility

FoSix:
## Description

Add code flexibility.

## Motivation and Context

Move away from hardcoded configuration in `main.tf`. This way this example can be used to deploy a particular code (or reference architecture) in different variations.

## How Has This Been Tested?

Deployed the example successfully.

## Types of changes

- refactor

## Checklist


examples/transit_vnet_common/example.tfvars
```diff
@@ -1,75 +1,207 @@
...
-olb_private_ip = "10.110.0.21"
 
-route_tables = {
-  private_route_table = {
-    routes = {
-      default = {
-        address_prefix         = "0.0.0.0/0"
-        next_hop_type          = "VirtualAppliance"
-        next_hop_in_ip_address = "10.110.0.21"
+# --- VNET PART --- #
```
migara, line 13:
```suggestion
# --- VNET CONFIGURATION --- #
```

examples/transit_vnet_common/versions.tf
```diff
@@ -11,5 +11,9 @@ terraform {
 }
 
 provider "azurerm" {
-  features {}
+  features {
+    resource_group {
+      prevent_deletion_if_contains_resources = false
```
migara, line 16: Shouldn't happen, but please verify the resource group will be retained if an existing resource group is used.

examples/transit_vnet_common/example.tfvars
```diff
@@ -1,75 +1,207 @@
...
 
-frontend_ips = {
-  "frontend01" = {
-    create_public_ip = true
-    rules = {
-      "balancessh" = {
-        protocol = "Tcp"
-        port     = 22
+
+# --- LOAD BALANCING PART --- #
```
migara, line 101:
```suggestion
# --- LOAD BALANCING CONFIGURATION --- #
```

examples/transit_vnet_common/example.tfvars
```diff
@@ -1,75 +1,207 @@
...
 
-common_vmseries_version = "9.1.3"
-common_vmseries_sku     = "bundle1"
-storage_account_name    = "pantfstorage"
-storage_share_name      = "bootstrapshare"
 
-files = {
-  "files/authcodes"    = "license/authcodes" # authcode is required only with common_vmseries_sku = "byol"
-  "files/init-cfg.txt" = "config/init-cfg.txt"
+# --- VMSERIES PART --- #
```
migara, line 143:
```suggestion
# --- VMSERIES CONFIGURATION --- #
```

examples/transit_vnet_common/example.tfvars
```diff
@@ -1,75 +1,207 @@
...
+      {
+        name               = "mgmt"
+        subnet_name        = "management"
+        private_ip_address = "10.0.0.10"
+        create_pip         = true
+      },
+      {
+        name                 = "private"
+        subnet_name          = "private"
+        backend_pool_lb_name = "lb-private"
```
migara, line 168: For clarity, let's use `backend_pool_name` instead. Otherwise, the user needs to have prior knowledge that [we are creating a backend pool with the load balancer name](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/9dfe470572d5b7b2e53cb8a30c43a116ea784870/modules/loadbalancer/main.tf#L119) and introduce a `backend_pool_name` after line #106

Not related to this, but please fix the [variable description](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/9dfe470572d5b7b2e53cb8a30c43a116ea784870/modules/loadbalancer/variables.tf#L154-L156) to indicate that we are defaulting to load balancer name if a name is not provided for the backend.

migara (approved):
Great work 👏🏽

examples/transit_vnet_common/versions.tf
FoSix: @migara this happens only in scenarios where AppInsights is deployed. Two resources are typically left. They are not created by Terraform. 

I also thought that since this is an example, not production code, we should just cleanup after our selves. Maybe a comment about that would nice.

examples/transit_vnet_common/example.tfvars
FoSix: well, you're right, this name is confusing, but this is not the backend name, it's actually the load balancer name. If you look [here](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/228-examples-refactor-transit_vnet_common/examples/transit_vnet_common/main.tf#L133) you'll see that we are using the `backend_pool_lb_name` property to target a specific load balancer and then we just access it's backend pool ID. Backend name in this situation i never used.

maybe we should change it to something like:

```suggestion
        load_balancer_name = "lb-private"
```

examples/transit_vnet_common/versions.tf
FoSix: > Shouldn't happen, but please verify the resource group will be retained if an existing resource group is used.

It will, this is used only when we create an RG with our code

PR #240, opened 2023-02-27
feat(module/loadbalancer): support for specifying backend port (#239)

### LINKED ISSUE #239: Support for frontend and backend ports in loadbalancer module
### Is your feature request related to a problem?

The current loadbalancer module does not allow users to specify a different port for the frontend and backend when creating loadbalacing rules.

### Describe the solution you'd like

Extend the module to accept a frontend and backend port for loadbalancing rules

### Describe alternatives you've considered

N/A

### Additional context

N/A

alperenkose:
## Description

Support for specifying a backend port in Load Balancer rules.

## Motivation and Context

Current Load Balancer module assumes frontend and backend ports to be the same in load balancing rules. This change enables to specify a different port for the backend pool where endpoints listen to.
Resolves #239 

## How Has This Been Tested?

Deployed a test spoke vnet which had vpc peering with Security VPC. Configured VMSeries firewalls to make destination NAT to application VMs on specified backend ports in the Load Balancer rules. 
Test scenario: LB frontend TCP port 80 - LB backend port 50080 - VMSeries DNAT 50080 to 80 - Application VM port 80

This feature required Network Security Group rule for untrust subnet to include LB backend port as destination port instead of the LB frontend port.

## Types of changes

## Checklist


modules/loadbalancer/outputs.tf
```diff
@@ -20,6 +20,7 @@ output "frontend_combined_rules" {
   ```
   {
     "frontend01-balancessh" = {
+      backend_port = null
```
FoSix, line 23: why do we explicitly set this to `null`

alperenkose: This is the output example. However in output_rules in locals we set backend_port to null if it's not provided. That's because the NSG rules are created with a for_each over output_rules in which we specify the destination port to match with the backend port if provided, or to match the front end port if not provided. Related line:

https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/239-support-for-frontend-and-backend-ports-in-loadbalancer-module/modules/loadbalancer/main.tf#L179

modules/loadbalancer/main.tf
```diff
@@ -135,7 +136,7 @@ resource "azurerm_lb_rule" "lb_rules" {
   backend_address_pool_ids = [azurerm_lb_backend_address_pool.lb_backend.id]
 
   protocol                       = each.value.rule.protocol
-  backend_port                   = each.value.rule.port
+  backend_port                   = try(each.value.rule.backend_port, each.value.rule.port)
```
sebastianczech, line 139: Behaviour of functions `try` and `coalesce` is different, especially when `each.value.rule.backend_port` will have value `""`. Should we here use `coalesce` instead of `try` ?

sebastianczech (approved):
LGTM :)

modules/loadbalancer/main.tf
alperenkose: Actually the `each.value.rule.backend_port` might not exist at all in the object. How about the following `coalesce(try(each.value.rule.backend_port, null), each.value.rule.port)` ?

alperenkose: Thanks!

modules/loadbalancer/outputs.tf
FoSix: 😳 you're right

modules/loadbalancer/main.tf
```diff
@@ -173,9 +174,9 @@ resource "azurerm_network_security_rule" "allow_inbound_ips" {
   direction                   = "Inbound"
   access                      = "Allow"
   protocol                    = title(replace(each.value.protocol, "all", "*"))
-  description                 = "Auto-generated for load balancer ${var.name} port ${each.value.protocol}/${each.value.port}: allowed inbound IP ranges"
+  description                 = "Auto-generated for load balancer ${var.name} port ${each.value.protocol}/${coalesce(each.value.backend_port, each.value.port)}: allowed inbound IP ranges"
```
FoSix, line 177:
```suggestion
  description                 = "Auto-generated for load balancer ${var.name} port ${each.value.protocol}/${coalesce(try(each.value.rule.backend_port, null), each.value.rule.port)}: allowed inbound IP ranges"
```

as `coalesce` does not work on non existing entities maybe we should use the same `coalesce(try())` combo also here

alperenkose: Actually it's already there. backend_port is set to null with a try condition in the output_rules, and we are looping over output_rules.

FoSix: 👍🏻

FoSix (approved):
👍🏻

PR #258, opened 2023-05-05
feat: checkov output pre commit

### LINKED ISSUE #200: Improve checkov output in pre-commit.
### Is your feature request related to a problem?

Right now we don't specify which test we want to check, and the result of that is whole errors which checkov makes are skipped because they touch something we don't need or wont to check.

### Describe the solution you'd like

We need to clarify which test we want to do with check and properly configure it.

horiagunica:
## Description

Implement Checkov fixes and remove global soft-fail.

This PR Closes #200 .

## Motivation and Context

Code security improvements.

## How Has This Been Tested?

Re-deploy all existing examples.

## Types of changes

## Checklist


examples/common_vmseries/main.tf
```diff
@@ -198,6 +198,7 @@ module "bootstrap" {
   name                   = each.value.name
   resource_group_name    = try(each.value.resource_group_name, local.resource_group.name)
   location               = var.location
+  storage_acl            = try(each.value.storage_acl, false)
```
FoSix, line 201: this property is missing in the bootstrap variable description

this applies to the `main.tf` files in the other (VMSeries based) examples

examples/dedicated_vmseries_and_autoscale/main.tf
```diff
@@ -11,7 +11,8 @@ resource "random_password" "this" {
 }
 
 locals {
-  vmseries_password = coalesce(var.vmseries_password, try(random_password.this[0].result, null))
+  vmseries_password               = coalesce(var.vmseries_password, try(random_password.this[0].result, null))
+  disable_password_authentication = local.vmseries_password == null ? true : false
```
FoSix, line 15: what do we need this one for?

modules/bootstrap/variables.tf
```diff
@@ -93,4 +93,43 @@ variable "tags" {
...
+  default     = 7
+  validation {
+    condition     = var.retention_policy_days > 0 && var.retention_policy_days < 365
+    error_message = "Enter a value between 1 and 365."
+  }
+}
+
+variable "storage_allow_inbound_public_ips" {
+  description = <<-EOF
+    List of IP CIDR ranges (like `["23.23.23.23"]`) that are allowed to access the storage.
```
FoSix, line 110:
```suggestion
    List of IP CIDR ranges (like `["23.23.23.23"]`) that are allowed to access the Storage Account.
```

modules/bootstrap/variables.tf
```diff
@@ -93,4 +93,43 @@ variable "tags" {
...
+}
+
+variable "storage_allow_vnet_subnets" {
+  description = <<-EOF
+  List of the allowed vnet subnets.
+  Note that this option requires network service endpoint enabled for Microsoft Storage for the specified subnets.
+  If you are using vnet module - set 'storage_private_access' to true for the specific subnet.
+  Example:
+  ```
+  [module.vnet.subnet_ids["subnet-mgmt"],module.vnet.subnet_ids["subnet-pub"],module.vnet.subnet_ids["subnet-priv"]]
```
FoSix, line 124:
```suggestion
  [
    module.vnet.subnet_ids["subnet-mgmt"],
    module.vnet.subnet_ids["subnet-pub"],
    module.vnet.subnet_ids["subnet-priv"]
  ]
```

modules/bootstrap/variables.tf
```diff
@@ -93,4 +93,43 @@ variable "tags" {
...
+  EOF
+  type        = list(string)
+  default     = null
+}
+
+variable "storage_allow_vnet_subnets" {
+  description = <<-EOF
+  List of the allowed vnet subnets.
+  Note that this option requires network service endpoint enabled for Microsoft Storage for the specified subnets.
+  If you are using vnet module - set 'storage_private_access' to true for the specific subnet.
```
FoSix, line 121:
```suggestion
  If you are using [vnet module](../vnet/README.md) - set `storage_private_access` to true for the specific subnet.
```

modules/bootstrap/variables.tf
```diff
@@ -93,4 +93,43 @@ variable "tags" {
...
+  ```
+  [module.vnet.subnet_ids["subnet-mgmt"],module.vnet.subnet_ids["subnet-pub"],module.vnet.subnet_ids["subnet-priv"]]
+  ```
+  EOF
+  type        = list(any)
+  default     = null
+}
+
+variable "storage_acl" {
+  description = "If `true`, storage account network rules will be activated with Deny default statement."
```
FoSix, line 132:
```suggestion
  description = "If `true`, storage account network rules will be activated with Deny as the default statement."
```

modules/virtual_machine/main.tf
```diff
@@ -77,12 +78,12 @@ resource "azurerm_virtual_machine" "this" {
...
     computer_name  = var.name
     admin_username = var.username
-    admin_password = var.password
+    admin_password = local.disable_password_authentication == true ? null : var.password
     custom_data    = var.custom_data
   }
 
   os_profile_linux_config {
-    disable_password_authentication = var.password == null ? true : false
+    disable_password_authentication = local.disable_password_authentication
```
FoSix, line 86: I would add here some `lifecycle.precondition` - it's possible now to set the `password` to `null` and not specify any SSH keys. At least one should be set.

this. can probably apply to the scale set as well

modules/vmss/main.tf
```diff
@@ -17,6 +17,13 @@ resource "azurerm_linux_virtual_machine_scale_set" "this" {
   zones                           = var.zones
   zone_balance                    = var.zone_balance
   provision_vm_agent              = false
+  dynamic "admin_ssh_key" {
+    for_each = var.ssh_key == null ? [] : [1]
```
FoSix, line 21: why do we rely here on a single key? and not on a list of keys like in the `vmseries` module?

FoSix (changes requested):
👍🏻 
some minor changes required

examples/dedicated_vmseries_and_autoscale/main.tf
horiagunica: That solves the issue in which checkov status that you must disable password auth for more security. Putting it under a condition eliminates the scan basically (because it's dependent on the input).

modules/virtual_machine/main.tf
horiagunica: Added lifecycle to both vmss and virtual machine in which you `var.password` and `var.ssh_keys` cannot be null at the same time :

```
lifecycle {
      precondition {
      condition = var.password != null || var.ssh_keys != null
      error_message = "Either password or ssh_keys must be set in order to have access to the device"
  }
}
```

modules/vmss/main.tf
horiagunica: Apologies - this was was left from a few months ago when my knowledge on the subiect was sparse - I updated it now.

examples/common_vmseries/main.tf
```diff
@@ -14,6 +14,12 @@ locals {
   vmseries_password = coalesce(var.vmseries_password, try(random_password.this[0].result, null))
 }
 
+# Obtain Public IP address of code deployment machine
+
+data "http" "this" {
```
FoSix, line 19: maybe let's add here a condition, I do not see much sense in running this source if we do not use full bootstrapping with acl. So either check if a bootstrap storage is defined, or even better, check if you have bootstrap storage with network ACL defined

examples/common_vmseries/variables.tf
```diff
@@ -272,6 +272,9 @@ variable "bootstrap_storage" {
   - `name` : name of the Storage Account. Please keep in mind that storage account name has to be globally unique. This name will not be prefixed with the value of `var.name_prefix`.
   - `create_storage_account` : (defaults to `true`) create or source (when `false`) an existing Storage Account.
   - `resource_group_name` : (defaults to `var.resource_group_name`) name of the Resource Group hosting the Storage Account (existing or newly created). The RG has to exist.
+  - `storage_acl` : (defaults to `false`) enables network ACLs on the Storage Account. If this is enabled - `storage_allow_vnet_subnets` and `storage_allow_inbound_public_ips` options become available. The ACL defaults to default `Deny`.
+  - `storage_allow_vnet_subnet_ids` : (defaults to `[]`) whitelist containing the allowed vnet and associated subnets that are allowed to access the Storage Account. Note that the respective subnets require `enable_storage_service_endpoint` set to `true` to work properly.
```
FoSix, line 276: these are not subnet IDs, but a map containing keys for vnet and subnet definitions. I would drop the `_ids` suffix
```suggestion
  - `storage_allow_vnet_subnets` : (defaults to `[]`) whitelist containing the allowed vnet and associated subnets that are allowed to access the Storage Account. Note that the respective subnets require `enable_storage_service_endpoint` set to `true` to work properly.
```

examples/common_vmseries/variables.tf
```diff
@@ -272,6 +272,9 @@ variable "bootstrap_storage" {
...
+  - `storage_allow_inbound_public_ips` : (defaults to `[]`) whitelist containing the allowed public IP subnets that can access the Storage Account. Note that the code automatically tried to query https://api.ipify.org to obtain the public IP address of the machine executing the code so that the bootstrap files are succuessfuly uploaded to the Storage Account.
```
FoSix, line 277:
```suggestion
  - `storage_allow_inbound_public_ips` : (defaults to `[]`) whitelist containing the allowed public IP subnets that can access the Storage Account. Note that the code automatically tries to query [https://api.ipify.org](https://api.ipify.org) to obtain the public IP address of the machine executing the code so that the bootstrap files can be successfully uploaded to the Storage Account.
```

examples/dedicated_vmseries/example.tfvars
```diff
@@ -151,7 +151,15 @@ bootstrap_storage = {
     name             = "xmplbootstrapdedicated"
     public_snet_key  = "public"
     private_snet_key = "private"
+    storage_acl      = true
     intranet_cidr    = "10.100.0.0/16"
+    storage_allow_vnet_subnet_ids = {
```
FoSix, line 156: this would have to follow the suggestion above
``` suggestion
storage_allow_vnet_subnets = {
```

examples/common_vmseries/main.tf
```diff
@@ -194,10 +200,13 @@ module "bootstrap" {
...
-  create_storage_account = try(each.value.create_storage, true)
-  name                   = each.value.name
-  resource_group_name    = try(each.value.resource_group_name, local.resource_group.name)
-  location               = var.location
+  create_storage_account           = try(each.value.create_storage, true)
+  name                             = each.value.name
+  resource_group_name              = try(each.value.resource_group_name, local.resource_group.name)
+  location                         = var.location
+  storage_acl                      = try(each.value.storage_acl, false)
+  storage_allow_vnet_subnet_ids    = try(flatten([for v in each.value.storage_allow_vnet_subnet_ids : [module.vnet[v.vnet_key].subnet_ids[v.subnet_key]]]), [])
```
FoSix, line 208: see suggestion below
```suggestion
  storage_allow_vnet_subnet_ids    = try(flatten([for v in each.value.storage_allow_vnet_subnets : [module.vnet[v.vnet_key].subnet_ids[v.subnet_key]]]), [])
```

FoSix (changes requested):

examples/common_vmseries/main.tf
horiagunica: Added a count condition to it to trigger only when the `bootstrap_storage` map variable is not empty and if the value of `storage_acl` is `true` :

```
data "http" "this" {
  count = length(var.bootstrap_storage) > 0 && contains([for v in values(var.bootstrap_storage) : v.storage_acl], true) ? 1 : 0
  url = "https://api.ipify.org"
}
```

FoSix (approved):
👍🏻

PR #273, opened 2023-06-19
feat: VNet peering support

horiagunica:
## Description

A module that allows VNet peering configuration between VNets within the same or different resource groups.

## Motivation and Context

No VNet peering module.

## How Has This Been Tested?

Tested within customer deployment.

## Types of changes

## Checklist


modules/vnet_peering/outputs.tf
FoSix, line 1: Just wondering if we need the `outputs` here or is this just a utility module that just does something and *exists*.

modules/vnet_peering/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+
+variable "name_prefix" {
+  description = "Prefix name appended to the peering names."
+  default     = ""
+  type        = string
+}
+
+variable "local_vnet_name" {
+  description = "Local VNET name."
+  default     = ""
```
FoSix, line 19: I would say: no defaults here. The VNET name is mandatory, otherwise the module does not make sense.

```suggestion
```

modules/vnet_peering/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+  type        = string
+}
+
+variable "peer_vnet_name" {
+  description = "Peer VNET name."
+  default     = ""
```
FoSix, line 25: same here

modules/vnet_peering/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+  type        = string
+}
+
+variable "local_allow_virtual_network_access" {
```
FoSix, line 29: this actually is applicable to all `local_` and `peer_` properties, this one and all the ones below: should we really duplicate them? This gives flexibility of course, but this is also a module that will be used with NGFWs, so traffic settings should be symetrical IMO.

horiagunica: For `allow_virtual_network_access` and `allow_forwarded_traffic` they could be the same but the other two actually require combinations for the NVA to work - for example if you want to use `use_remote_gateways` in one VNET - you must have `use_remote_gateways` + `allow_gateway_transit` in the other one - otherwise it will not work .

It's a bit more variables - true - but at least it offers you full flexibility .

modules/vnet_peering/outputs.tf
horiagunica: It's just something to be there so we can have an outputs file - we may change the output itself . I doubt there are any resources in which we would use outputs - but I didn't want to not have one either.

FoSix (approved):
👍🏻

modules/vnet_peering/main.tf
```diff
@@ -0,0 +1,31 @@
+data "azurerm_virtual_network" "local" {
```
michalbil, line 1: We do output virtual_network_id in the VNet module, so we could allow to either provide the IDs directly (thus skip extra datasource read) or get ones based on RG/VNet name as it's done right now.

modules/vnet_peering/main.tf
```diff
@@ -0,0 +1,31 @@
...
+  resource_group_name = var.local_resource_group_name
+}
+
+data "azurerm_virtual_network" "peer" {
+  name                = var.peer_vnet_name
+  resource_group_name = var.peer_resource_group_name
+}
+
+resource "azurerm_virtual_network_peering" "local" {
+  name                         = "${var.name_prefix}${var.local_vnet_name}-to-${var.peer_vnet_name}"
```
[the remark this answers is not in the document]
horiagunica: Agreed! Added!

modules/vnet_peering/outputs.tf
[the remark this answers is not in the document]
horiagunica: Added!

modules/vnet_peering/main.tf
FoSix: that would be nice, but what would happen if you specify both? we would need some additional logic to cover that. 

And that would be two additional variables, making this small module quite complex. I'm not saying no, I'm far from that, but maybe we should also rethink how to define variables so that they are more grouped? readable?

group them into maps? one for local vnet, one for remote? just and idea 🤔

horiagunica: I refactored the code using maps - but it does seem that we have some checks failing for version 1.2 when the validate job is running . I believe the optional value inside the object is the issue :
```
  type = object({
    vnet_name                    = string
    resource_group_name          = string
    name                         = optional(string)
    allow_virtual_network_access = bool
    allow_forwarded_traffic      = bool
    allow_gateway_transit        = bool
    use_remote_gateways          = bool
  })
```

I did find something regarding this topic here : https://medium.com/@apr_1985/terraform-optional-attributes-and-defaults-d73643185341

horiagunica: After discussion with @FoSix - we will use `map(any)` for now until we introduce a refactoring of the `map(any)` variables.

modules/vnet_peering/variables.tf
```diff
@@ -0,0 +1,45 @@
+variable "name_prefix" {
+  description = "Prefix name appended to the peering names."
+  default     = ""
+  type        = string
+}
+
+variable "local_peer_config" {
+  description = <<-EOF
+  A map that contains the local peer configuration.
+  Mandatory Values:
```
FoSix, line 10: since [this](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/82370696752ed913e697945e114e4d0830fe4fa6/examples/dedicated_vmseries_and_autoscale/variables.tf#L258) format is more or less approved, can you align variables to match it? It will be less work when we will refactor the other modules.

modules/vnet_peering/variables.tf
```diff
@@ -0,0 +1,35 @@
...
+  description = "Prefix name appended to the peering names."
+  default     = ""
+  type        = string
+}
+
+variable "local_peer_config" {
+  description = <<-EOF
+  A map that contains the local peer configuration.
+  Mandatory Values: 
+  - `vnet_name`                   - (`string`, required) : the local peer VNET name.
```
FoSix, line 11:
```suggestion
  - `vnet_name`                   - (`string`, required) the local peer VNET name.
```

modules/vnet_peering/main.tf
```diff
@@ -0,0 +1,31 @@
...
+  resource_group_name = var.local_peer_config.resource_group_name
+}
+
+data "azurerm_virtual_network" "remote_peer" {
+  name                = var.remote_peer_config.vnet_name
+  resource_group_name = var.remote_peer_config.resource_group_name
+}
+
+resource "azurerm_virtual_network_peering" "local" {
+  name                         = try(var.local_peer_config.name, "${var.name_prefix}${var.local_peer_config.vnet_name}-to-${var.remote_peer_config.vnet_name}")
```
FoSix, line 12:
```suggestion
  name                         = try(var.local_peer_config.name, "${var.local_peer_config.vnet_name}-to-${var.remote_peer_config.vnet_name}")
```

two reasons for that change:
- when working with existing resource name (like `vnet_name`, it has to be a full name of an existing resource, with prefixes and everything, as we use it in a datasource) we would already have a prefix starting the peering name, so for `hub-` prefix this would give us: `hub-hub-vnet-hub-othervnet` <- does not make sense
- this is not a resource, meaning you cannot see it in a resource group (if I'm not wrong here), hence it's name should be descriptive, but does not need to have a prefix IMO, you will see it inside a VNET in the peerings list

modules/vnet_peering/main.tf
```diff
@@ -0,0 +1,31 @@
...
+  virtual_network_name         = var.local_peer_config.vnet_name
+  remote_virtual_network_id    = data.azurerm_virtual_network.remote_peer.id
+  allow_virtual_network_access = try(var.local_peer_config.allow_virtual_network_access, true)
+  allow_forwarded_traffic      = try(var.local_peer_config.allow_forwarded_traffic, true)
+  allow_gateway_transit        = try(var.local_peer_config.allow_gateway_transit, false)
+  use_remote_gateways          = try(var.local_peer_config.use_remote_gateways, false)
+}
+
+resource "azurerm_virtual_network_peering" "remote" {
+  name                         = try(var.remote_peer_config.name, "${var.name_prefix}${var.remote_peer_config.vnet_name}-to-${var.local_peer_config.vnet_name}")
```
FoSix, line 23: [this comment](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/pull/273/files#r1267689419) applies also here

horiagunica: Done! 👍

PR #276, opened 2023-06-22
ci: add terraform 1.5 into equation

FoSix:
## Description

Extend existing CIs to run checks also using Terraform 1.5

## Motivation and Context

Terraform 1.5 released.

This PR is also replacing https://api.ipify.org endpoint which seems to be buggy and unresponsive, with https://ifconfig.me/ip

## How Has This Been Tested?

Run CI, results are [here](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/actions/runs/5345677808?pr=276)

## Types of changes

## Checklist


examples/standalone_panorama/main.tf
```diff
@@ -88,4 +88,6 @@ module "panorama" {
 
   tags       = var.tags
   depends_on = [module.vnet]
-}
\ No newline at end of file
+}
+
+# test
```
pimielowski, line 93: Some debug?

.github/workflows/release_ci.yml
```diff
@@ -20,5 +20,4 @@ jobs:
     with:
       cloud: azure
       max_parallel: 10
-      tf_version: 1.2 1.3 1.4
```
pimielowski, line 23: Why you remove the tf_version here?

pimielowski (changes requested):
Could you double check this deleted tf_version if it is ok?

.github/workflows/release_ci.yml
FoSix: Following the discussion yesterday on planning - we will run `apply` only on the latest TF version, if you do not specify any here it will automatically pick up `latest`, if you put here any version, `1.5` in example, you will have to update the code each time something new is released

examples/standalone_panorama/main.tf
FoSix: yeah, good catch :smile:

pimielowski (approved):
LGTM!

.github/workflows/release_ci.yml
pimielowski: Got it!

Issue #257, opened 2023-04-26
unify the format of variables documentation


FoSix:
this is related mainly to `examples`, but module might require review as well
proposed format - VMSS documentation in autoscaling example

michalbil:
First proposal from my side - use hyphen instead of colons to separate property name and description (such an approach is a standard for example in TF resources docs) and remove extra spaces from (type|required) - pipe alone is a good enough separator imo.

```hcl
description = <<-EOF
Some general variable description.

Some longer description if required: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. 

If this is a map, information about keys and values meaning should be included.

If values are nested maps, a list of keys/properties could follow:
  - `name_of_property` - (`variable_type`|`required`) some short description.
  - `example_string_required` - (`string`|`required`) for required properties
  - `example_bool_optional` - (`bool`|`false`) when property is not required and we force a default value in the example, possibly overwrting the module default value
  - `example_string_optional` - (`string`|`module default`) when property is not required and we rely on module's defaults, this implicitly states that the variable defined in a module has `nullable` set to `false`
  - `example_number_optional` - (`number`|`cloud default`) when property is not required and we rely on the defaults set by the cloud API

Example

example blocks could follow here
EOF
```

| Name | Description | Type | Default | Required |
|------|-------------|------|---------|:--------:|
[...]

PR #285, opened 2023-07-05
DO NOT MERGE unify the format of variables documentation

FoSix:
This PR is here only to highlight differences when witching to the new variables documentation format as described in #257


alperenkose:
I think the Reference section was better organized with the table format but I also think that it looks bad on pan.dev since it doesn't fit within the available width. And having the reference sections organized in the navigation pane in pan.dev is a nice thing to have. Considering both having the documentation formatted in markdown doc looks better to me.

sebastianczech:
Great job 👍 
I like that approach to have 1 sentence for each variable in table and then details below.
I have 2 questions:
- what is the reason to treat `.header.md` as hidden file, not just `header.md` ?
- do we have possibility to add syntax colouring for examples embedded in variables description ?

FoSix:
@sebastianczech 

> what is the reason to treat `.header.md` as hidden file, not just `header.md` ?

the header file is just a part of the Readme, not a *standalone* document. The dot and idea of *hiding* it is just to emphasise that fact

> do we have possibility to add syntax colouring for examples embedded in variables description ?

Sure, just add language specs in the variable description, like [here](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/a83b7c13bf42d63e3057cc48ac0d0d6729c0337f/modules/vnet/variables.tf#L70)
We didn't use that in the pass as table formatting didn't support that correctly. With document format the description is copied 1:1, hence all MD features should be supported

PR #287, opened 2023-07-07
ci: initroduce chatops

FoSix:
# ChatOPS - on demand testing

## Why

Sometimes there is a need to test something more/else that what is available in `PR CI`. For example: `PR CI` does not deploy any infrastructure, for complex changes it might be a good idea to deploy the infra before the `Release CI` (which runs deployment on all examples before releasing a version).

To make this possible a chatops-like CI is introduced. It reacts on `/test` command in a PR's comment and triggers a [`_tf_plan_apply`](https://github.com/PaloAltoNetworks/terraform-modules-vmseries-ci-workflows/blob/main/.github/workflows/_tf_plan_apply.yml) workflow with similar parameters to those passed by `Release CI`. This basically means a possibility to deploy infrastructure during a PR.

## How

To run the deployment add a comment to a PR with a command, one line. The command for the moment is `/test`.

After the command parameters can be placed, space delimited list. For the moment there are three parameters: 

- `paths` - (required) a space delimited list of paths that should be tested. This can be a single path as well
- `tf_version` - (optional) a space delimited list of `Terraform versions`. When omited the latest version will be used.
- `do_apply` - (optional) `true` or `false` (default) - triggers actual deployment, or just planning

  **NOTE** \
  The TF versions specified have to be supported by the modules.

## Examples

- run tests on a single path, TF versions 1.2 and 1.4:

  ```bash
  /test paths="examples/standalone_vmseries" tf_version="1.2 1.4"
  ```

- run tests on two paths, latest TF version, with deployment:

  ```bash
  /test paths="examples/standalone_vmseries examples/standalone_panorama" do_apply=true
  ```

## The Flow

1. A command is issued as a comment to a PR
2. The [`ChatOPS dispatcher`](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/chatops/.github/workflows/chatops.yml) workflow kicks in
   1. it analyses the comment
   2. tries to run the actual test CI
   3. if it succeeds, the comment is updated with 🚀  and 👀 reactions - this is an indicator that the test workflow is running 
   4. if it fails, a comment is updated with a failure reason and the reaction is changed to 👎🏻 
       image

3. The [`ChatOPS tester`](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/chatops/.github/workflows/test-command.yml) workflow is dispatched
   1. it updates the comment with a link to the test job
   2. it triggers the [`_tf_plan_apply`](https://github.com/PaloAltoNetworks/terraform-modules-vmseries-ci-workflows/blob/main/.github/workflows/_tf_plan_apply.yml) reusable workflow passing all required information.
   3. when the reusable workflow is finished the comment is updated with the test results. For `success` the reaction is changed to 👍🏻  and 👎🏻 otherwise.
     image

## Testing

To test it one can use a [copy of the Azure repo](https://github.com/PaloAltoNetworks/vm-series-gh-actions/pull/71) (internal), as the chatops workflows need to be on the `main` branch to trigger.

### DIFF
--- a/.github/workflows/chatops.yml
+++ b/.github/workflows/chatops.yml
@@ -1,0 +1,57 @@
+name: ChatOPS dispatcher
+run-name: "ChatOPS bot for PR - (#${{ github.event.issue.number }}) ${{ github.event.issue.title }}"
+
+permissions:
+  contents: read
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  dispatch:
+    name: Dispatch a test job
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - name: get PR head branch
+        uses: actions/github-script@v6
+        id: pr
+        with:
+          result-encoding: string
+          script: |
+            let pr = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number,
+            })
+            console.log(pr.data.head.ref)
+            return pr.data.head.ref
+            
+      - name: "dispatch test command on branch: ${{ steps.pr.outputs.result }}"
+        id: scd
+        uses: peter-evans/slash-command-dispatch@v3
+        with:
+          token: ${{ secrets.CHATOPS }}
+          issue-type: pull-request
+          dispatch-type: workflow
+          permission: write
+          commands: |
+            test
+          static-args: |
+            comment-id=${{ github.event.comment.id }}
+            pr-id=${{ github.event.issue.number }}
+            pr-title=${{ github.event.issue.title }}
+            branch=${{ steps.pr.outputs.result }}
+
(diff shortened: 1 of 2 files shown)


.github/workflows/chatops.yml
```diff
@@ -0,0 +1,57 @@
...
+            return pr.data.head.ref
+            
+      - name: "dispatch test command on branch: ${{ steps.pr.outputs.result }}"
+        id: scd
+        uses: peter-evans/slash-command-dispatch@v3
+        with:
+          token: ${{ secrets.CHATOPS }}
+          issue-type: pull-request
+          dispatch-type: workflow
+          permission: maintain
```
FoSix, line 40: @migara - pushed the required access to run workflows to `maintain`

FoSix:
obsolete, changes will be introduced elsewhere

Issue #280, opened 2023-06-30
unify all examples to a single codebase


FoSix:

alperenkose:
We maybe also add enable_zones conditional to appgw modules in the examples which would make it dependent on this flag like the other modules in the examples.

```
zones              = var.enable_zones ? try(each.value.zones, null) : null
```

FoSix:
superseded to some extent by #307

Issue #279, opened 2023-06-30
Refactor map(any) variables


horiagunica:
### Is your feature request related to a problem?

Currently map variables have `any` structure which does not constrain the values within the map.

### Describe the solution you'd like

Use `object({})` and define variable constraints - for example :

```
  type = object({
    vnet_name                    = string
    resource_group_name          = string
    name                         = optional(string)
    allow_virtual_network_access = bool
    allow_forwarded_traffic      = bool
    allow_gateway_transit        = bool
    use_remote_gateways          = bool
  })
```

### Additional context

If we are going to use this implementation - we would require to bump the TF version to 1.3 minimum so that we can use `optional` values within the map object without constraints : https://medium.com/@apr_1985/terraform-optional-attributes-and-defaults-d73643185341

PR #299, opened 2023-07-24
feat(examples/common_vmseries_and_autoscale): example for SWFW Hub

alperenkose:
## Description

Common VMSeries with Autoscaling example for SWFW Hub

## How Has This Been Tested?

Deployed with spoke vnet and tested inbound/outbound/east-west traffic flows.

## Types of changes

## Checklist


examples/common_vmseries_and_autoscale/variables.tf
```diff
@@ -0,0 +1,362 @@
...
+  - `scalein_config` : (map|`{}`) scale in configuration, same properties supported as for `scaleout_config`
+
+  Example, no auto scaling:
+
+  ```
+  {
+  "vmss" = {
+    name              = "ngfw-vmss"
+    vnet_key          = "transit"
+    bootstrap_options = "type=dhcp-client"
```
FoSix, line 305: there's a bug in the dedicated vmss: `"type=dhcp"`, maybe we could use this PR to fix it there

examples/common_vmseries_and_autoscale/example.tfvars
```diff
@@ -0,0 +1,247 @@
+# --- GENERAL --- #
+location            = "North Europe"
+resource_group_name = "autoscale-common"
+name_prefix         = "example-"
+tags = {
+  "CreatedBy"   = "Palo Alto Networks"
+  "CreatedWith" = "Terraform"
+}
+enable_zones = false
```
FoSix, line 9: dedicated vmss is non-zonal. Maybe we could make this example zone-enabled?

FoSix (approved):
👍🏻 very nice, already mergeable in this form

examples/common_vmseries_and_autoscale/variables.tf
alperenkose: Yes saw that, I was going to open a separate PR for that but let me update this then.

examples/common_vmseries_and_autoscale/main.tf
```diff
@@ -0,0 +1,239 @@
...
+  location            = var.location
+  subnet_id           = module.vnet[each.value.vnet_key].subnet_ids[each.value.subnet_key]
+
+  managed_identities = try(each.value.managed_identities, null)
+  waf_enabled        = try(each.value.waf_enabled, false)
+  capacity           = try(each.value.capacity, null)
+  capacity_min       = try(each.value.capacity_min, null)
+  capacity_max       = try(each.value.capacity_max, null)
+  enable_http2       = try(each.value.enable_http2, null)
+  zones              = var.enable_zones ? try(each.value.zones, null) : null
```
alperenkose, line 155: Shall we add this `var.enable_zones` condition to other examples as well? It doesn't check for the enable_zones variable in others.

examples/common_vmseries_and_autoscale/example.tfvars
alperenkose: converted the example to zone-enabled but there is a couple of things to note; since NATGWs are zonal resources it's removed from the example and the outbound internet traffic is provided from public IP addresses assigned to public interfaces of firewalls in the VMSS. 
Another point is since we don't have a NATGW, I needed to enable public IPs on management interfaces as well, otherwise it wouldn't be possible for the firewalls to bootstrap with publicly accessible Panorama or do their initial licensing with Palo Alto update servers since they won't have internet access. 
We might add a note for this on README perhaps; mentioning that public IP addresses on management interfaces can be removed after the deployment and all management traffic can be forwarded to the firewalls' private interfaces to get NATed to the internet (after the firewalls get configured to NAT this traffic). Any other ideas?

FoSix: > Another point is since we don't have a NATGW, I needed to enable public IPs on management interfaces as well, otherwise it wouldn't be possible for the firewalls to bootstrap with publicly accessible Panorama or do their initial licensing with Palo Alto update servers since they won't have internet access.

that's not entirely true. A VM with just private IP and default NSG rules (or no NSG) can access the internet, it's not accessible from the internet. So you don't need public IPs on management to contact a publicly available Panorama (which probably will never happen) or to download updates.
In our examples we do not block the default outbound route, so we do not need public IPs for that.

examples/common_vmseries_and_autoscale/main.tf
FoSix: I would keep keep it as it is and start to create a list of things to refactor. We already have #280 and #279 that will make a huge change in the code. Maybe we can move all three to a single issue?

examples/common_vmseries/main.tf
```diff
@@ -115,7 +115,6 @@ module "load_balancer" {
       subnet_id                = try(module.vnet[v.vnet_key].subnet_ids[v.subnet_key], null)
       in_rules                 = try(v.in_rules, {})
       out_rules                = try(v.out_rules, {})
-      zones                    = var.enable_zones ? try(v.zones, null) : null # For the regions without AZ support.
```
FoSix, line 118: why did you remove it?

examples/common_vmseries_and_autoscale/example.tfvars
```diff
@@ -0,0 +1,243 @@
...
+    name              = "common-vmss"
+    vnet_key          = "transit"
+    zones             = ["1", "2", "3"]
+    bootstrap_options = "type=dhcp-client"
+
+    interfaces = [
+      {
+        name       = "management"
+        subnet_key = "management"
+        create_pip = true
```
FoSix, line 205: as discussed in some other place, we do not need this

examples/common_vmseries_and_autoscale/example.tfvars
```diff
@@ -0,0 +1,243 @@
...
+        name              = "private"
+        subnet_key        = "private"
+        load_balancer_key = "private"
+      },
+      {
+        name                    = "public"
+        subnet_key              = "public"
+        load_balancer_key       = "public"
+        application_gateway_key = "public"
+        create_pip              = true
```
FoSix, line 217: we could also skip this and use outbound rules in public LB?

examples/dedicated_vmseries/main.tf
```diff
@@ -115,7 +115,6 @@ module "load_balancer" {
       subnet_id                = try(module.vnet[v.vnet_key].subnet_ids[v.subnet_key], null)
       in_rules                 = try(v.in_rules, {})
       out_rules                = try(v.out_rules, {})
-      zones                    = var.enable_zones ? try(v.zones, null) : null # For the regions without AZ support.
```
FoSix, line 118: same here, why this got removed?

examples/dedicated_vmseries_and_autoscale/main.tf
```diff
@@ -110,7 +110,6 @@ module "load_balancer" {
       subnet_id                = try(module.vnet[v.vnet_key].subnet_ids[v.subnet_key], null)
       in_rules                 = try(v.in_rules, {})
       out_rules                = try(v.out_rules, {})
-      zones                    = var.enable_zones ? try(v.zones, null) : null # For the regions without AZ support.
```
FoSix, line 113: and again

examples/standalone_vmseries/main.tf
```diff
@@ -115,7 +115,6 @@ module "load_balancer" {
       subnet_id                = try(module.vnet[v.vnet_key].subnet_ids[v.subnet_key], null)
       in_rules                 = try(v.in_rules, {})
       out_rules                = try(v.out_rules, {})
-      zones                    = var.enable_zones ? try(v.zones, null) : null # For the regions without AZ support.
```
FoSix, line 118: same

FoSix (changes requested):

examples/dedicated_vmseries/main.tf
alperenkose: This variable is not there anymore in load balancer module. avzones variable within the load balancer is used instead.

examples/dedicated_vmseries_and_autoscale/main.tf
alperenkose: Same with others.

examples/common_vmseries_and_autoscale/example.tfvars
alperenkose: As we don't have LB outbound rules in reference architectures we decided to keep it the same.

examples/common_vmseries_and_autoscale/example.tfvars
alperenkose: After testing and troubleshooting with public IPs removed from management interfaces, we have found out that default outbound access is blocked if VMs are attached to a Load Balancer or they have any additional interface with a public IP assigned. We have both so relying on the default outbound access won't be possible. 

And since NATGWs are zonal, this leaves us with 2 feasible options I believe:
* Assigning public IP to both public and management interfaces
* Creating a separate backend pool for the management interfaces in order to create outbound rules from them (but this is problematic as well since we don’t support creating separate backend pools for different interfaces on our current LB module)

As a note if we assign public IPs to the management interfaces and the user wants to remove them we can put a comment on README saying that public IP addresses on management interfaces can be removed after the deployment with management traffic forwarded to the firewalls’ private interfaces (private LB) to get NATed to the internet (after the firewalls get configured to NAT this traffic).

alperenkose: It is agreed to keep public IP addresses on management interfaces for the simplicity of the example and put a disclaimer on how to avoid it.

examples/common_vmseries_and_autoscale/main.tf
alperenkose: Converted it back to match with the other examples. Put a comment on #280 for this.

examples/common_vmseries_and_autoscale/example.tfvars
alperenkose: Agreed to keep it. Details on the other thread.

FoSix (approved):

migara (approved):

PR #316, opened 2023-09-13
ci: Use a GitHub App for chatops token

migara:
## Description

A personal access token is not suitable to be used with the peter-evans/slash-command-dispatch@v3 action.

This will introduce a step to generate a token via a GitHub App in place of a PAT.

## Types of changes

## Checklist

### DIFF
--- a/.github/workflows/chatops.yml
+++ b/.github/workflows/chatops.yml
@@ -36,7 +36,7 @@ jobs:
       
       - name: Generate GitHub token
         id: generate-token
-        uses: tibdex/github-app-token@v1
+        uses: tibdex/github-app-token@v2
         with:
           app_id: ${{ secrets.CHATOPS_APP_ID }}
           private_key: ${{ secrets.CHATOPS_APP_PRIVATE_KEY }}
--- a/.github/workflows/chatops.yml
+++ b/.github/workflows/chatops.yml
@@ -40,7 +40,8 @@ jobs:
         with:
           app_id: ${{ secrets.CHATOPS_APP_ID }}
           private_key: ${{ secrets.CHATOPS_APP_PRIVATE_KEY }}
-          installation_id: ${{ secrets.CHATOPS_APP_INSTALLATION_ID }}
+          installation_retrieval_mode: id
+          installation_retrieval_payload: ${{ secrets.CHATOPS_APP_INSTALLATION_ID }}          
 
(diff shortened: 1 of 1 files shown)


FoSix (approved):
👍🏻

.github/workflows/chatops.yml
```diff
@@ -33,12 +33,20 @@ jobs:
             })
             console.log(pr.data.head.ref)
             return pr.data.head.ref
-            
+      
+      - name: Generate GitHub token
+        id: generate-token
+        uses: tibdex/github-app-token@v1
```
sebastianczech, line 39: Maybe we should go with [`v2`](https://github.com/tibdex/github-app-token/releases/tag/v2.0.0), especially if it revokes the created token at the end of the job with a [post script](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runspost).

sebastianczech (approved):

.github/workflows/chatops.yml
migara, line 39:
```suggestion
        uses: tibdex/github-app-token@v2
```

.github/workflows/chatops.yml
```diff
@@ -33,12 +33,20 @@ jobs:
...
             return pr.data.head.ref
-            
+      
+      - name: Generate GitHub token
+        id: generate-token
+        uses: tibdex/github-app-token@v2
+        with:
+          app_id: ${{ secrets.CHATOPS_APP_ID }}
+          private_key: ${{ secrets.CHATOPS_APP_PRIVATE_KEY }}
+          installation_id: ${{ secrets.CHATOPS_APP_INSTALLATION_ID }}
```
sebastianczech, line 43: In `v2` we don't have `installation_id`, so maybe we should configure it in below way:

```suggestion
          installation_retrieval_mode: id
          installation_retrieval_payload: ${{ secrets.CHATOPS_APP_INSTALLATION_ID }}          
```

or maybe we should use `installation_retrieval_mode` as `organization` as defined in https://github.com/tibdex/github-app-token/blob/v2.0.0/action.yml ?

sebastianczech (approved):

.github/workflows/chatops.yml
migara: Good catch

PR #318, opened 2023-09-15
refactor!: Changed avset name to key

acelebanski:
## Description

Minor changes to examples supporting Availability Sets. 

Changed the var name from `availability_set_name` to `availability_set_key` because in fact a key is used.
Changed the var name from `availability_set` to `availability_sets` to be compliant with naming convention (plural).

## Motivation and Context

It doesn't solve any issues, it's just a cosmetic change so the var names reflect reality, naming convention and the "key convention".
However, these examples will not be compatible with tfvars files from previous releases, hence "breaking change".

## How Has This Been Tested?

Example with Availability Sets after changing the var name has been deployed in the lab and in customer's environment.

## Types of changes

## Checklist


michalbil (approved):

sebastianczech (approved):
👍

michalbil:
Would be great to rename `var.availability_set` to `var.availability_sets` at some point - to better reflect it's contents and align with convention used for other variables.

acelebanski:
As per Michal's suggestion - I changed the name of variable `availability_set` to `availability_sets` in the same examples (to comply with naming convention) as it makes sense to change both of these vars' names at the same time.

pimielowski (approved):
LGTM 👍

Issue #307, opened 2023-08-29
refactor modules


FoSix:
An issue tracker for modules (and example) refactor:

- [ ] add flexible names to all resources created by a module
- [ ] move from `type = any` to defining the actual maps with `optional` values
- [ ] #257
- [ ] #280

alperenkose:
**A NOTE**: I think we should add these guidelines to somewhere like CONTRIBUTING before closing this issue; in order to follow the same conventions in the future.

PR #339, opened 2023-10-06
refactor(module/virtual_network_gateway): Refactor module for VNG

sebastianczech:
## Description

PR delivers changes for VNG:
- `README.md` adjusted to new style with additional header file
- minimum Terraform version set to 1.3
- refactor variables:
   - descriptions
   - types with optionals
   - default values
   - validations
- simplify `main.tf` by removing not required `try`
- remove name generation logic from the module

## Motivation and Context

#307 , #333

## How Has This Been Tested?

I checked that module using example included in PR.

## Types of changes

## Checklist


acelebanski:
Hi Sebastian, shouldn't you make `name` a mandatory property everywhere in the module and get rid of prefix/suffix etc.? In order to move the entire name generation logic to example?

sebastianczech:
@acelebanski thank you for feedback. I removed `name_prefix` and `name_suffix`. All resources have `name` attribute.

modules/virtual_network_gateway/variables.tf
```diff
@@ -8,46 +15,16 @@ variable "location" {
   type        = string
```
FoSix, line 15: the description says `load balancer`, should be VNG

modules/virtual_network_gateway/variables.tf
```diff
@@ -8,46 +15,16 @@ variable "location" {
...
-  ie: for zone-redundant with 3 availability zone in current region value will be:
-  ```["1","2","3"]```
-  EOF
-  default     = []
-  type        = list(string)
-}
-
+# Virtual Network Gateway
 variable "type" {
   description = "The type of the Virtual Network Gateway. Valid options are Vpn or ExpressRoute. Changing the type forces a new resource to be created"
```
FoSix, line 26: maybe lets use `EOF` notation for wrapping descriptions?

modules/virtual_network_gateway/variables.tf
```diff
@@ -8,46 +15,16 @@ variable "location" {
...
+  default     = "Vpn"
   type        = string
```
FoSix, line 28: what if a null value is assigned? can we live with that? or should we do `nullable=false`?

modules/virtual_network_gateway/variables.tf
FoSix: Same applies to other multi-sentence descriptions

modules/virtual_network_gateway/variables.tf
FoSix: same applies to all variables with default values

modules/virtual_network_gateway/variables.tf
```diff
@@ -108,22 +86,37 @@ variable "generation" {
 
 variable "private_ip_address_enabled" {
   description = "Should private IP be enabled on this gateway for connections?"
```
FoSix, line 88: This should start with one sentence description, like: `Controls whether the private IP is enabled on the gateway.`

modules/virtual_network_gateway/variables.tf
```diff
@@ -108,22 +86,37 @@ variable "generation" {
...
+  default     = false
   type        = bool
 }
 
+variable "avzones" {
```
FoSix, line 93: how about chainging to `zones` ?
```suggestion
variable "zones" {
```

modules/virtual_network_gateway/variables.tf
```diff
@@ -108,22 +86,37 @@ variable "generation" {
...
+  description = <<-EOF
+  After provider version 3.x you need to specify in which availability zone(s) you want to place IP.
+
+  For zone-redundant with 3 availability zone in current region value will be:
+  ```["1","2","3"]```
+  EOF
+  default     = []
+  type        = list(string)
```
FoSix, line 101: I'm wondering shouldn't we add validation here, either no zones, or all 3? Is that an option? I'm guessing this reflects the PIP as well

modules/virtual_network_gateway/variables.tf
```diff
@@ -136,59 +129,95 @@ variable "ip_configuration" {
...
     }
   ]
-
+  ```
   EOF
-  type        = list(any)
+  default     = []
+  nullable    = false
+  type = list(object({
+    name                          = optional(string, "vnetGatewayConfig")
```
FoSix, line 137: can we have two configs with the same name? if no, I would drop the default value and leave the name property as required + a validation that checks if the name property is unique

modules/virtual_network_gateway/variables.tf
```diff
@@ -136,59 +129,95 @@ variable "ip_configuration" {
...
+    create_public_ip              = bool
+    public_ip_name                = optional(string)
+    private_ip_address_allocation = optional(string, "Dynamic")
+    public_ip_standard_sku        = optional(bool, false)
```
FoSix, line 141: do we even consider doing basic?

modules/virtual_network_gateway/variables.tf
```diff
@@ -136,59 +129,95 @@ variable "ip_configuration" {
...
-
+  ```
   EOF
-  type        = list(any)
+  default     = []
+  nullable    = false
+  type = list(object({
+    name                          = optional(string, "vnetGatewayConfig")
+    create_public_ip              = bool
+    public_ip_name                = optional(string)
```
FoSix, line 139: when looking at the code it looks like the name property is required, not optional, see `main.tf`, lines `101`&`108` and `118`&`120`

modules/virtual_network_gateway/variables.tf
```diff
@@ -136,59 +129,95 @@ variable "ip_configuration" {
...
-  - revoked_certificate - one or more revoked_certificate blocks which are defined below.
-  - radius_server_address - the address of the Radius server.
-  - radius_server_secret - the secret used by the Radius server.
-  - vpn_client_protocols - list of the protocols supported by the vpn client. The supported values are SSTP, IkeV2 and OpenVPN. Values SSTP and IkeV2 are incompatible with the use of aad_tenant, aad_audience and aad_issuer.
-  - vpn_auth_types - list of the vpn authentication types for the virtual network gateway. The supported values are AAD, Radius and Certificate.
+  VPN client configurations (IPSec point-to-site connections).
+
+  List of available attributes of each VPN client configurations:
+  - `address_space`           - (`string`, required) the address space out of which IP addresses for vpn clients will be taken. You can provide more than one address space, e.g. in CIDR notation.
+  - `aad_tenant`              - (`string`, optional) AzureAD Tenant URL
```
FoSix, line 152: I would suggest the following whenever we have optional value, even if it's `null`
```suggestion
  - `aad_tenant`              - (`string`, optional, defaults to `null`) AzureAD Tenant URL
```

modules/virtual_network_gateway/variables.tf
```diff
@@ -200,41 +229,57 @@ variable "local_bgp_settings" {
       }
     }
   }
-
+  ```
   EOF
-  type        = any
+  type = object({
```
FoSix, line 234: all the properties are optional but the variable is required?

modules/virtual_network_gateway/variables.tf
```diff
@@ -200,41 +229,57 @@ variable "local_bgp_settings" {
...
+  Every object in the map contains attributes:
+  - local_ng_name           - (`string`, required) the name of the local network gateway.
+  - connection_name         - (`string`, required) the name of the virtual network gateway connection.
+  - remote_bgp_settings     - (`list`, optional) block containing Local Network Gateway's BGP speaker settings:
+    - asn                   - (`string`, required) the BGP speaker's ASN.
+    - bgp_peering_address   - (`string`, required) the BGP peering address and BGP identifier of this BGP speaker.
+    - peer_weight           - (`number`, optional) the weight added to routes learned from this BGP speaker.
+  - gateway_address         - (`string`, optional) the gateway IP address to connect with.
+  - address_space           - (`list`, optional) the list of string CIDRs representing the address spaces the gateway exposes.
+  - custom_bgp_addresses    - (`list`, optional) Border Gateway Protocol custom IP Addresses, which can only be used on IPSec / active-active connections. Object contains 2 attributes:
```
FoSix, line 272: try maybe wrapping manually lines longer than 130 chars

modules/virtual_network_gateway/variables.tf
```diff
@@ -293,36 +338,69 @@ variable "local_network_gateways" {
...
+  }
 }
 
 variable "connection_mode" {
-  description = "Connection mode to use. Possible values are Default, InitiatorOnly and ResponderOnly. Defaults to Default. Changing this value will force a resource to be created."
+  description = "Connection mode to use."
+  default     = "Default"
   type        = string
+  validation {
+    condition     = contains(["Default", "InitiatorOnly", "ResponderOnly"], var.connection_mode)
```
FoSix, line 376: this validation could fail if you assign `null`

FoSix: try adding `nullable=false`

modules/virtual_network_gateway/main.tf
```diff
@@ -2,78 +2,78 @@
...
-      aad_tenant    = try(vpn_client_configuration.value.aad_tenant, null)
-      aad_audience  = try(vpn_client_configuration.value.aad_audience, null)
-      aad_issuer    = try(vpn_client_configuration.value.aad_issuer, null)
+      address_space = vpn_client_configuration.value.address_space
+      aad_tenant    = vpn_client_configuration.value.aad_tenant
+      aad_audience  = vpn_client_configuration.value.aad_audience
+      aad_issuer    = vpn_client_configuration.value.aad_issuer
       dynamic "root_certificate" {
-        for_each = try(vpn_client_configuration.value.root_certificate, null) != null ? { for t in vpn_client_configuration.value.root_certificate : t.name => t } : {}
+        for_each = vpn_client_configuration.value.root_certificate != null ? { for t in vpn_client_configuration.value.root_certificate : t.name => t } : {}
```
FoSix, line 37: instead of doing a condition against `null` you can use coalesce, like this:
```suggestion
        for_each = { for t in coalesce(vpn_client_configuration.value.root_certificate, {}) : t.name => t }
```

modules/virtual_network_gateway/main.tf
```diff
@@ -2,78 +2,78 @@
...
+          default_addresses     = peering_addresses.value.default_addresses
         }
       }
-      peer_weight = try(bgp_settings.value.peer_weight, null)
+      peer_weight = bgp_settings.value.peer_weight
     }
   }
 
   dynamic "custom_route" {
     for_each = var.custom_route
```
FoSix, line 74: this will brake when you assign null to `var.custom_route`, see comments in `variables.tf` about `nullable`

modules/virtual_network_gateway/main.tf
```diff
@@ -97,22 +97,25 @@ resource "azurerm_virtual_network_gateway" "this" {
...
   if ip_configuration.create_public_ip }
 
   resource_group_name = var.resource_group_name
   location            = var.location
-  name                = "${var.name_prefix}pip-vgw${var.name_suffix}-${each.key}"
+  name                = each.value.name
 
-  allocation_method = each.value ? "Static" : "Dynamic"
-  zones             = var.enable_zones ? var.avzones : null
+  allocation_method = each.value.public_ip_standard_sku ? "Static" : "Dynamic"
```
FoSix, line 110: Dynamic allocation has two properties:
1. it does not need NSG for incoming internet traffic
2. IP is being rotated when the service becomes offline

For VPN configs I would say we should only do Static + NSG - to be sure that the IP is always the same and to control the incoming traffic

modules/virtual_network_gateway/main.tf
```diff
@@ -97,22 +97,25 @@ resource "azurerm_virtual_network_gateway" "this" {
...
+  zones             = try(length(var.avzones) > 0, false) ? var.avzones : null
```
FoSix, line 111: maybe instead of doing this complex condition play around with `nullable` and/or default values, for example, if we set `active-active` to `false` by default, set `zones` to `null` by default. next to that to verifications:
- validation in `zones` variable: either `null` or a valid list of zones
- precondition in PIP resource, when `active-active` is true, `zones` cannot be `null`

modules/virtual_network_gateway/main.tf
```diff
@@ -97,22 +97,25 @@ resource "azurerm_virtual_network_gateway" "this" {
 # https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/public_ip
 resource "azurerm_public_ip" "this" {
```
FoSix, line 98: I would place this in the top, I know it does not make more difference for TF, but it makes the code more self-documenting

modules/virtual_network_gateway/main.tf
```diff
@@ -97,22 +97,25 @@ resource "azurerm_virtual_network_gateway" "this" {
...
   resource_group_name = var.resource_group_name
   location            = var.location
-  name                = "${var.name_prefix}pip-vgw${var.name_suffix}-${each.key}"
+  name                = each.value.name
 
-  allocation_method = each.value ? "Static" : "Dynamic"
-  zones             = var.enable_zones ? var.avzones : null
+  allocation_method = each.value.public_ip_standard_sku ? "Static" : "Dynamic"
+  zones             = try(length(var.avzones) > 0, false) ? var.avzones : null
   tags              = var.tags
```
FoSix, line 112: place tags as the bottom of a resource - that's the convention in azure resource we have for a long time

modules/virtual_network_gateway/main.tf
```diff
@@ -144,38 +147,38 @@ resource "azurerm_local_network_gateway" "this" {
...
 
-  enable_bgp                     = try(var.enable_bgp, false)
-  local_azure_ip_address_enabled = try(var.private_ip_address_enabled, true)
+  enable_bgp                     = var.enable_bgp
+  local_azure_ip_address_enabled = var.private_ip_address_enabled
   shared_key                     = var.ipsec_shared_key
 
   dynamic "custom_bgp_addresses" {
-    for_each = try(each.value.custom_bgp_addresses, {})
+    for_each = each.value.custom_bgp_addresses
```
FoSix, line 163: this is defined as an optional list, w/o any default value, hence `null` - this will fail -> add a default value of `[]` to type

modules/virtual_network_gateway/main.tf
```diff
@@ -122,18 +125,18 @@ data "azurerm_public_ip" "exists" {
...
-  name                = "${var.name_prefix}lgw-vgw${var.name_suffix}-${each.value.name}"
+  name                = each.value.local_ng_name
   resource_group_name = var.resource_group_name
   location            = var.location
   gateway_address     = each.value.gateway_address
-  address_space       = try(each.value.address_space, [])
+  address_space       = each.value.address_space
 
   dynamic "bgp_settings" {
     for_each = each.value.remote_bgp_settings
```
FoSix, line 135: this is defined as an optional list, w/o any default value, hence null - this will fail -> add a default value of [] to type

FoSix (changes requested):
Check the:

- EOF notation 
- nullable for vars with defaults
- 1st description sentence should be a brief variable description - it will land in the top-most inputs table, examples: default_local_network_gateway_id, edge_zone, etc
- validations, like the one in line 40: add `nullable=false` and you can skip `coalesce`
- in main.tf, when you use nullable combined with default values you can skip coalesce in cases like the one in line 83 - biggest advantage is that you have the default value defined in one place - the variable definition, and not scattered around the code

modules/virtual_network_gateway/variables.tf
sebastianczech: fixed

modules/virtual_network_gateway/variables.tf
sebastianczech: fixed

modules/virtual_network_gateway/variables.tf
sebastianczech: fixed

modules/virtual_network_gateway/variables.tf
sebastianczech: Thanks, I've added `nullable=false` (in other places too)

modules/virtual_network_gateway/variables.tf
sebastianczech: ok

modules/virtual_network_gateway/main.tf
sebastianczech: thanks, changed

modules/virtual_network_gateway/variables.tf
sebastianczech: added, thanks

modules/virtual_network_gateway/main.tf
sebastianczech: fixed, thanks

modules/virtual_network_gateway/main.tf
sebastianczech: ok, fixed

modules/virtual_network_gateway/main.tf
sebastianczech: moved

modules/virtual_network_gateway/.header.md
```diff
@@ -0,0 +1,7 @@
+# Palo Alto Networks Virtual Network Gateway Module for Azure
+
+A terraform module for deploying a Virtual Network Gateway and its components required for the VM-Series firewalls in Azure.
+
+## Usage
+
+For usage refer to variables description, which include example for complex map of objects.
```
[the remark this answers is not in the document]
sebastianczech: yes, I've added an example

modules/virtual_network_gateway/variables.tf
sebastianczech: resolved

modules/virtual_network_gateway/variables.tf
sebastianczech: yes, you are right, fixed

modules/virtual_network_gateway/main.tf
sebastianczech: added

modules/virtual_network_gateway/main.tf
sebastianczech: added

modules/virtual_network_gateway/variables.tf
sebastianczech: ok, fixed this and the others optionals too

modules/virtual_network_gateway/variables.tf
sebastianczech: done

modules/virtual_network_gateway/variables.tf
sebastianczech: I propose to leave that flexibility to choose e.g. for the lab do we need standard ?

modules/virtual_network_gateway/variables.tf
sebastianczech: ok, I added

modules/virtual_network_gateway/variables.tf
sebastianczech: you are right, fixed

modules/virtual_network_gateway/main.tf
sebastianczech: I propose to leave that flexibility to choose e.g. for the lab cannot we use dynamic ?

modules/virtual_network_gateway/main.tf
sebastianczech: I added

modules/virtual_network_gateway/main.tf
sebastianczech: According to internal discussion, I defined allocation method for public IP as `Static` and SKU `Standard`.

modules/virtual_network_gateway/main.tf
```diff
@@ -1,86 +1,119 @@
+# https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/public_ip
+resource "azurerm_public_ip" "this" {
+  for_each = { for ip_configuration in var.ip_configuration : ip_configuration.name => ip_configuration.name if ip_configuration.create_public_ip }
```
FoSix, line 3: I'm not saying you code is wrong :) just an idea I'd like to share:
```hcl
toset([ for ipc in var.ip_configuration : ipc.name if ipc.create_public_ip ])
```

modules/virtual_network_gateway/main.tf
```diff
@@ -1,86 +1,119 @@
...
+
+  resource_group_name = var.resource_group_name
+  location            = var.location
+  name                = each.value
+
+  allocation_method = "Static"
+  sku               = "Standard"
+  zones             = try(length(var.zones) > 0, false) ? var.zones : null
```
FoSix, line 11: cannot this be just `var.zones`? default is null and everything else is governed by the validation block

modules/virtual_network_gateway/main.tf
```diff
@@ -144,38 +153,38 @@ resource "azurerm_local_network_gateway" "this" {
...
   shared_key                     = var.ipsec_shared_key
 
   dynamic "custom_bgp_addresses" {
-    for_each = try(each.value.custom_bgp_addresses, {})
+    for_each = each.value.custom_bgp_addresses
     content {
-      primary   = try(var.azure_bgp_peers_addresses[custom_bgp_addresses.value.primary], null)
-      secondary = try(var.azure_bgp_peers_addresses[custom_bgp_addresses.value.secondary], null)
+      primary   = var.azure_bgp_peers_addresses[custom_bgp_addresses.value.primary]
+      secondary = custom_bgp_addresses.value.secondary != null ? var.azure_bgp_peers_addresses[custom_bgp_addresses.value.secondary] : null
```
FoSix, line 172: `try` would be shorter

modules/virtual_network_gateway/main.tf
```diff
@@ -1,86 +1,119 @@
+# https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/public_ip
```
FoSix, line 1: what do you think about adding this as a standard? so to all `main.tf` files?

modules/virtual_network_gateway/variables.tf
```diff
@@ -91,39 +89,62 @@ variable "edge_zone" {
...
 
-  - name - name of the IP configuration
-  - create_public_ip - boolean value, true if public IP needs to be created
-  - public_ip_name - name of the public IP resource used, when there is no need to create new one
-  - private_ip_address_allocation - defines how the private IP address of the gateways virtual interface is assigned. Valid options are Static or Dynamic. Defaults to Dynamic.
-  - public_ip_standard_sku - defaults to `false`, when set to `true` creates a Standard SKU, statically allocated public IP, otherwise it will be a Basic/Dynamic one.
-  - subnet_id - the ID of the gateway subnet of a virtual network in which the virtual network gateway will be created.
+  - `name`                          - (`string`, required) name of the IP configuration
+  - `create_public_ip`              - (`bool`, required) - true if public IP needs to be created
+  - `public_ip_name`                - (`string`, required) name of the public IP resource used, when there is no need to create new one
```
FoSix, line 141: you're not consistens when it comes to line wrapping :) , see lines: 141-143, 195, 197, 202, 259, 297, 320, 323 and 324

modules/virtual_network_gateway/main.tf
sebastianczech: done

modules/virtual_network_gateway/main.tf
sebastianczech: done

modules/virtual_network_gateway/main.tf
sebastianczech: I agree :) I was adding this in comments to have quick link to the documentation while developing the module :)

modules/virtual_network_gateway/main.tf
sebastianczech: done

modules/virtual_network_gateway/variables.tf
sebastianczech: fixed

FoSix (approved):

PR #346, opened 2023-10-13
refactor(module/application_insights)!: module refactor and adjusted examples

### LINKED ISSUE #326: application_insights refactor

alperenkose:
## Description

Closes #326 

This branch is based on #307.

## Types of changes

## Checklist


examples/common_vmseries/variables.tf
```diff
@@ -229,40 +273,33 @@ variable "availability_set" {
...
-
-  * when the value is set to `null` (default) no AI is created
-  * when the value is a map containing `name` key (other keys are optional) a single AI instance will be created under the name that is the value of the `name` key
-  * when the value is an empty map or a map w/o the `name` key, an AI instance per each VMSeries VM will be created. All instances will share the same configuration. All instances will have names corresponding to their VM name.
-
-  Names for all AI instances are prefixed with `var.name_prefix`.
+  A map defining Azure Application Insights.
 
-  Properties supported (for details on each property see [modules documentation](../../modules/application_insights/README.md)):
+  For detailed documentation on each property refer to [module documentation](../../modules/application_insights/README.md)
```
FoSix, line 278: there are only 4 properties here, I would avoid pointing the user to module's documentation if it's not really necensary, like when you omit some nested props to avoid duplicating documentation.

examples/common_vmseries/variables.tf
```diff
@@ -229,40 +273,33 @@ variable "availability_set" {
...
-  }
   ```
   EOF
-  default     = null
-  type        = map(string)
+  default     = {}
+  type = map(object({
+    name                      = string
+    workspace_name            = string
+    workspace_sku             = optional(string, "PerGB2018")
```
FoSix, line 300: see comment around `workspace_sku` variable, if we need to keep the default value in the module's variable definition, add `nullable` to the definition, skip the default value here. This way you control the default value only in one place and you do not overwrite it in the example code.

examples/common_vmseries/variables.tf
```diff
@@ -229,40 +273,33 @@ variable "availability_set" {
...
+  For detailed documentation on each property refer to [module documentation](../../modules/application_insights/README.md)
 
-  - `name` : (optional, string) a name of a single AI instance
-  - `workspace_mode` : (optional, bool) defaults to `true`, use AI Workspace mode instead of the Classical (deprecated)
-  - `workspace_name` : (optional, string) defaults to AI name suffixed with `-wrkspc`, name of the Log Analytics Workspace created when AI is deployed in Workspace mode
-  - `workspace_sku` : (optional, string) defaults to PerGB2018, SKU used by WAL, see module documentation for details
-  - `metrics_retention_in_days` : (optional, number) defaults to current Azure default value, see module documentation for details
+  - `name`                      - (`string`, required) name of the Application Insights instance.
+  - `workspace_name`            - (`string`, required) name of the Log Analytics Workspace to be created together with the AI instance.
+  - `workspace_sku`             - (`string`, optional, defaults "PerGB2018") SKU used by WAL, for details see [Application Insights module documentation](../../modules/application_insights/README.md#workspace_sku).
```
FoSix, line 282: if we need to keep the default value in the module, set the parentised text to soemthing like `optional, default to module defaults`

FoSix (changes requested):
some minor thing to check

rebase with 307 before mergining

FoSix:
this will be refactored with the #335

PR #352, opened 2023-10-26
refactor(module/loadbalancer): separate fip and rules

FoSix:
An example how separation of FIPs and rules would look like


sebastianczech:
For me it's cleaner approach, we don't have so much locals, validations can be splited into separated variables etc. :)

sebastianczech (approved):
👍

acelebanski:
Hey Łukasz, are you still planning to prepare more complex examples to compare old vs new way of doing things?

Issue #364, opened 2023-11-27
[Bug Report] Panorama Module "boot_diagnostics.0.storage_uri" is required, but no definition was found.


PenDrag0n7:
### Describe the bug

I set the boot_diagnostic_storage_uri to null but it still asking for the url.

```
│ Error: Missing required argument
│
│   with module.panorama["panorama-ae"].azurerm_virtual_machine.panorama,
│   on .terraform\modules\panorama\modules\panorama\main.tf line 79, in resource "azurerm_virtual_machine" "panorama":
│   79:     storage_uri = var.boot_diagnostic_storage_uri
│
│ The argument "boot_diagnostics.0.storage_uri" is required, but no definition was found.
╵
```

### Module Version

1.2.2

### Terraform version

 v1.5.6

### Expected behavior

Should allow to skip the need to set storage account for bootdiag

### Anything else to add?

Problematic code in panorama module

  boot_diagnostics {
    enabled     = var.boot_diagnostic_storage_uri != null ? true : false
    storage_uri = var.boot_diagnostic_storage_uri
  }

FoSix:
@PenDrag0n7 - thanks for pointing that out. I think we missed that because everyone is used to a workaround you can find in the [panorama example](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/main/examples/standalone_panorama/main.tf#L73).

But you're right, the code should work with just the defaults. A fix is coming.

PR #365, opened 2023-12-05
fix(examples/standalone_panorama): fix to be deployable with the defaults

FoSix:
## Description

A fix for how we are handling the bootstrap storage in the panorama module. 

## Motivation and Context

As pointed out by @PenDrag0n7 (#364), the module is not usable with the default values. The problematic part was the condition around bootstrap storage uri parameter that did not allow a default null.

## How Has This Been Tested?

A workaround found in the Panorama example was removed and example was deployed.

## Types of changes

## Checklist

### DIFF
--- a/examples/standalone_panorama/main.tf
+++ b/examples/standalone_panorama/main.tf
@@ -60,17 +60,16 @@ module "panorama" {
 
   for_each = var.panoramas
 
-  name                        = "${var.name_prefix}${each.value.name}"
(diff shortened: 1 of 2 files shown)


FoSix:
/idempotence paths="examples/standalone_panorama" tf_version="1.2 1.3 1.4 1.5 1.6"

pimielowski (approved):

PR #371, opened 2023-12-19
refactor: bootstrap and vmseries

FoSix:


examples/vmseries/main.tf
```diff
@@ -0,0 +1,339 @@
...
+  backend_pool_ids = { for v in local.full_nics_list_flat : "${v.vm}-${v.nic_name}" => v }
+}
+
+resource "azurerm_network_interface_backend_address_pool_association" "this" {
+  for_each = local.backend_pool_ids
+
+  backend_address_pool_id = module.load_balancer[each.value.lb_key].backend_pool_id
+  ip_configuration_name   = module.vmseries[each.value.vm].interfaces[each.value.nic_name].ip_configuration[0].name
+  network_interface_id    = module.vmseries[each.value.vm].interfaces[each.value.nic_name].id
+}
```
acelebanski, line 339: Still not sure about adding this resource into the glue code, hope we can discuss on call.

acelebanski (commented):
Left a few comments, mainly regarding **vmseries** module.

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,232 @@
...
+  type = list(object({
+    name                     = string
+    subnet_id                = string
+    create_public_ip         = optional(bool, false)
+    public_ip_name           = optional(string)
+    public_ip_resource_group = optional(string)
+    private_ip_address       = optional(string)
+    lb_backend_pool_id       = optional(string)
+  }))
+  # validation ip name required when create = true
```
acelebanski, line 231: You could make it like this I believe.


```hcl
  validation {
    condition = (var.interfaces.create_public_ip == true && length(var.interfaces.public_ip_name) > 0
      ) || (var.interfaces.create_public_ip == false && length(var.interfaces.public_ip_name) > 0
    ) || (var.interfaces.create_public_ip == false && var.interfaces.public_ip_name == null)
    error_message = "The `public_ip_name` property is required when `create_public_ip` is `true`."
  }
```

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,232 @@
...
+  validation {
+    condition     = contains(["Standard_LRS", "StandardSSD_LRS", "Premium_LRS"], var.virtual_machine.disk_type)
+    error_message = "The `disk_type` property can be one of: `Standard_LRS`, `StandardSSD_LRS` or `Premium_LRS`."
+  }
+  validation {
+    condition     = contains(["SystemAssigned", "UserAssigned", "SystemAssigned, UserAssigned"], var.virtual_machine.identity_type)
+    error_message = "The `identity_type` property can be one of \"SystemAssigned\", \"UserAssigned\" or \"SystemAssigned, UserAssigned\"."
+  }
+  validation {
+    condition     = var.virtual_machine.identity_type == "SystemAssigned" ? length(var.virtual_machine.identity_ids) == 0 : length(var.virtual_machine.identity_ids) >= 0
```
acelebanski, line 160: In false case, shouldn't `var.virtual_machine.identity_ids` length be higher than 0 rather than higher or equal?

acelebanski (commented):
Some more comments on variables validation.

modules/bootstrap/main.tf
```diff
@@ -1,55 +1,34 @@
...
   count = var.create_storage_account ? 1 : 0
 
   name                     = var.name
   location                 = var.location
   resource_group_name      = var.resource_group_name
-  min_tls_version          = var.min_tls_version
+  min_tls_version          = var.storage_network_security.min_tls_version
   account_replication_type = "LRS"
   account_tier             = "Standard"
+  account_kind             = "StorageV2"
```
sebastianczech, line 11: What do you think about defining new variable for `account_kind` with default value `StorageV2` ?

modules/bootstrap/versions.tf
```diff
@@ -1,13 +1,9 @@
 terraform {
-  required_version = ">= 1.2, < 2.0"
+  required_version = ">= 1.5, < 2.0"
```
sebastianczech, line 2: Not `required_version = ">= 1.3, < 2.0"` ? Are we starting from TF 1.5 ?

modules/vmseries/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 1.2, < 2.0"
+  required_version = ">= 1.5, < 2.0"
```
sebastianczech, line 2: The same as in previous one - shouldn't we use `required_version = ">= 1.3, < 2.0"` ?

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,232 @@
...
-variable "diagnostics_storage_uri" {
-  description = "The storage account's blob endpoint to hold diagnostic files."
-  default     = null
-  type        = string
+  type = list(object({
+    name                     = string
+    subnet_id                = string
+    create_public_ip         = optional(bool, false)
+    public_ip_name           = optional(string)
+    public_ip_resource_group = optional(string)
```
sebastianczech, line 227: Is it name of the resource group, not key , yes ? Then maybe we should use name  ` public_ip_resource_group_name` instead of ` public_ip_resource_group` ?

modules/vmseries/main.tf
```diff
@@ -1,128 +1,122 @@
...
+
+resource "azurerm_network_interface_backend_address_pool_association" "this" {
+  for_each = { for v in var.interfaces : v.name => v.lb_backend_pool_id if v.lb_backend_pool_id != null }
+
+  backend_address_pool_id = each.value
+  ip_configuration_name   = azurerm_network_interface.this[each.key].ip_configuration[0].name
+  network_interface_id    = azurerm_network_interface.this[each.key].id
+
+  depends_on = [
+    azurerm_network_interface.this,
```
sebastianczech, line 119: Do we need directly pass dependency `azurerm_network_interface.this` if we have ` ip_configuration_name   = azurerm_network_interface.this[each.key].ip_configuration[0].name` ?

modules/bootstrap/main.tf
```diff
@@ -61,47 +40,109 @@ locals {
...
+
+  inverted_filenames = {
+    for k, _ in var.file_shares : k => merge(local.bootstrap_filenames[k], local.inverted_files[k])
+  }
+
+  # invert local.filenames map
+  filenames = {
+    for k, _ in var.file_shares : k => { for _k, _v in local.inverted_filenames[k] : _v => _k }
+  }
+  filenames_across_fileshares_flat = flatten([
```
sebastianczech, line 112: Can we in comments describe in few sentences what is the purpose of having inverted files , then flat filenames across file shares etc. ? It's not clear while looking on that code at the beginning.

sebastianczech (commented):
As always great job 👍

modules/bootstrap/main.tf
FoSix: possible, but most of the settings other than `StorageV2` requier `Premium` tier, so we would have to parametrize both

modules/vmseries/main.tf
```diff
@@ -1,128 +1,122 @@
...
 
   location            = var.location
   resource_group_name = var.resource_group_name
-  name                = "${each.value.name}-pip"
+  name                = each.value.public_ip_name
   allocation_method   = "Static"
   sku                 = "Standard"
-  zones               = var.enable_zones ? var.avzones : null
-  tags                = try(each.value.tags, var.tags)
+  zones               = var.virtual_machine.zone != null ? [var.virtual_machine.zone] : null
```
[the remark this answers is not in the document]
FoSix: yes, done on purpose, making a zone redundand IP for a zonal vm does not really make sense, and was introduced due to some limitation in the 1st versions of the 3.x provider

modules/vmseries/main.tf
FoSix: yes, this was causing problem with terraform destroy see [here](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/commit/d7f7116bc027b1608e0032c88cd6ee0bb125e9ba)

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,232 @@
...
+  type = object({
+    size                         = optional(string, "Standard_D3_v2")
+    bootstrap_options            = optional(string)
+    zone                         = string
+    disk_type                    = optional(string, "StandardSSD_LRS")
+    disk_name                    = string
+    avset_id                     = optional(string)
+    accelerated_networking       = optional(bool, true)
+    encryption_at_host_enabled   = optional(bool)
+    proximity_placement_group_id = optional(string)
```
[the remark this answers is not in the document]
FoSix: yeah, this should go away

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,232 @@
...
-  - `tags`                     - (optional|map) Tags to assign to the interface and public IP (if created). Overrides contents of `tags` variable.
+  Following configuration options are available:
+
+  - `name`                      - (`string`, required) the interface name
+  - `subnet_id`                 - (`string`, required) ID of an existing subnet to create the interface in
+  - `private_ip_address`        - (`string`, optional, defaults to `null`) static private IP to assign to the interface. When
+                                  skipped Azure will assign one dynamically.
+                                
+      Keep in mind that a dynamic IP is guarantied not to change as long as the VM is running. Any stop/deallocate/restart
+      operation might cause the IP to change.
```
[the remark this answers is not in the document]
FoSix: as we discussed during the call sometimes it makes sens to put it into a separate paragraph, especially if the description is long, or contains codeblocks. But in this case you are right

modules/vmseries/variables.tf
FoSix: well you are probably right, the problem is that we already use this convention in other modules, like [`load balancer`](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/a63213f80532ba637c7d785f53d70c642a0ee59a/modules/loadbalancer/variables.tf#L72)

so if we decide to change it, it has to be everywhere

and besides, if a property holds a key, it should be prefixed with a `_key`, like [here for example](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/ff2b20e5f0770267b3ab6bccd8b65a682f26fc43/examples/vmseries/variables.tf#L376)

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,232 @@
...
+  EOF
+  type = object({
+    size                         = optional(string, "Standard_D3_v2")
+    bootstrap_options            = optional(string)
+    zone                         = string
+    disk_type                    = optional(string, "StandardSSD_LRS")
+    disk_name                    = string
+    avset_id                     = optional(string)
+    accelerated_networking       = optional(bool, true)
+    encryption_at_host_enabled   = optional(bool)
```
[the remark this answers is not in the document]
FoSix: fixed

modules/vmseries/variables.tf
FoSix: this was fixed in all examples and modules, from now on all properties that hold a name are suffixed with `_name`

FoSix:
/idempotence paths="examples/common_vmseries examples/dedicated_vmseries" tf_version="1.5. 1.6"

acelebanski (commented):
Left comments on the modules code.

examples/common_vmseries/example.tfvars
```diff
@@ -22,7 +22,7 @@ vnets = {
             direction                  = "Inbound"
             access                     = "Allow"
             protocol                   = "Tcp"
-            source_address_prefixes    = ["1.2.3.4"]
+            source_address_prefixes    = ["134.238.135.14", "134.238.135.140"]
```
acelebanski, line 25: We sure we want to add our public IPs into the examples? I'm more into adding `0.0.0.0/0` with a comment for the user to update this one with his own values.

examples/standalone_vmseries/variables.tf
acelebanski, line 1: I believe we don't need `enable_zones` variable anymore.

acelebanski, line 1: Lines 213-218, bad indentation.

modules/bootstrap/variables.tf
```diff
@@ -1,12 +1,7 @@
...
-  description = "If `true`, create a Storage Account."
-  default     = true
-  type        = bool
-}
-
 variable "name" {
   description = <<-EOF
-  Name of the Storage Account, either a new or an existing one (depending on the value of `create_storage_account`).
+  Name of the Storage Account.
+  Either a new or an existing one (depending on the value of `create_storage_account`).
```
alperenkose, line 4:
```suggestion
  Either a new or an existing one (depending on the value of `storage_account.create`).
```

modules/bootstrap/.header.md
```diff
@@ -0,0 +1,126 @@
...
+The sample below shows how to source an existing Storage Account with an existing File Share.
+
+Please **note** that we will also skip bootstrap package folder structure creation. The sourced File Share should have this folder
+structure already present.
+
+```hcl
+module "existing_storage" {
+  source = "../../modules/bootstrap"
+
+  create_storage_account = false
```
alperenkose, line 101:
```suggestion
  storage_account = {
    create = false
  }  
```

modules/loadbalancer/variables.tf
```diff
@@ -69,7 +69,7 @@ variable "frontend_ips" {
   - `create_public_ip`          - (`bool`, optional, defaults to `false`) when set to `true` a new public IP will be
                                   created, otherwise an existing resource will be used;
                                   in both cases the name of the resource is controled by `public_ip_name` property
-  - `public_ip_resource_group`  - (`string`, optional, defaults to the Load Balancer's RG) name of a Resource Group
+  - `public_ip_resource_group_name`  - (`string`, optional, defaults to the Load Balancer's RG) name of a Resource Group
```
alperenkose, line 72: Shall we always keep the same empty space between key - description for all properties or is it different on purpose here?

modules/vmseries/main.tf
```diff
@@ -1,128 +1,129 @@
...
   # After converting to azurerm_linux_virtual_machine, an empty block boot_diagnostics {} will use managed storage. Want.
   # 2.36 in required_providers per https://github.com/terraform-providers/terraform-provider-azurerm/pull/8917
-  dynamic "boot_diagnostics" {
-    for_each = var.diagnostics_storage_uri != null ? ["one"] : []
-
-    content {
-      enabled     = true
-      storage_uri = var.diagnostics_storage_uri
-    }
+  boot_diagnostics {
```
alperenkose, line 107: since we are converting to `azurerm_linux_virtual_machine` can we support the managed storage option as in the code comments above?

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,243 @@
...
+  - `publisher`           - (`string`, optional, defaults to `paloaltonetworks`) the Azure Publisher identifier for a image
+                            which should be deployed
+  - `offer`               - (`string`, optional, defaults to `vmseries-flex`) the Azure Offer identifier corresponding to a
+                            published image
+  - `sku`                 - (`string`, optional, defaults to `byol`) VMSeries SKU; list available with
+                            `az vm image list -o table --all --publisher paloaltonetworks`
+  - `enable_marketplace_plan` - (`bool`, optional, defaults to `true`) when set to `true` accepts the license for an offer/plan
+                                on Azure Market Place
+  - `custom_id`         - (`string`, optional, defaults to `null`) absolute ID of your own custom PanOS image to be used for
+                                creating new Virtual Machines
```
alperenkose, line 66: different description indentations

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,243 @@
...
+
+  This map contains basic, as well as some optional Firewall parameters. Both types contain sane defaults.
+  Nevertheless they should be at least reviewed to meet deployment requirements.
+
+  List of either required or important properties: 
+
+  - `size`              - (`string`, optional, defaults to `Standard_D3_v2`) Azure VM size (type). Consult the *VM-Series
+                          Deployment Guide* as only a few selected sizes are supported
+  - `zone`              - (`number`, ??????????) Availability Zone to place the VM in, `null` value means a non-zonal deployment
+                          this Firewall will be created, explicit `null` means non-zonal deployment
```
alperenkose, line 100: this line looks like a duplicate description.

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,243 @@
...
+
+      ```hcl
+      bootstrap_options = "type=dhcp-client;panorama-server=1.2.3.4"
+      ```
+
+      For more details on bootstrapping [see documentation](https://docs.paloaltonetworks.com/vm-series/10-2/vm-series-deployment/bootstrap-the-vm-series-firewall/create-the-init-cfgtxt-file/init-cfgtxt-file-components).
+
+  List of other, optional properties: 
+
+  - `avset_key`                      - (`string`, optional, default to `null) identifier of the Availability Set to use
```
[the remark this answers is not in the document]
alperenkose: missing closing quotes for `null`

examples/virtual_network_gateway/main.tf
alperenkose, line 1: Sorry I probably missed it. Why do we remove some examples like VNG?

alperenkose (changes requested):

examples/virtual_network_gateway/main.tf
migara: Example of VNG on its own doesn't add any value and just clutter up the /examples

examples/common_vmseries/example.tfvars
FoSix: yeah, you are right, although I was not paying that much attention to examples yet

examples/virtual_network_gateway/main.tf
FoSix: we needed the example per module in the early development phase. we're almost done with modules refactor, hence I figured we do not need these examples any more.

modules/bootstrap/variables.tf
```diff
@@ -18,139 +13,211 @@ variable "name" {
...
+      calculated automatically. It's only required for files modified/created by Terraform. You can use `md5` or `filemd5`
+      Terraform functions to calculate MD5 hashes dynamically.
+
+      Keys in this map are local paths, variables - MD5 hashes. For files for which you would like to provide MD5 hashes, 
+      keys in this map should match keys in `bootstrap_files` property.
+
+
+  Additionally you can override the default `quota` and `access_tier` properties per File Share (same restrictions apply):
+
+  - `quota`       - (`number`, optional, defaults to `10`) maximum size of a File Share in GB, a value between 1 and
```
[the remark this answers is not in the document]
FoSix: my mistake, this defaults to `file_shares_configuration` values, where default is `10`

modules/bootstrap/variables.tf
```diff
@@ -18,139 +13,211 @@ variable "name" {
...
+                    5120 (5TB)
+  - `access_tier` - (`string`, optional, defaults to `Cool`) access tier for a File Share, can be one of: "Cool", "Hot",
```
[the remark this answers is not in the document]
FoSix: same here

modules/loadbalancer/variables.tf
FoSix: this is just a mistake, I was fixing the `_name` suffix with `sed` or something.

modules/vmseries/main.tf
```diff
@@ -1,128 +1,129 @@
...
+# https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/data-sources/public_ip
 data "azurerm_public_ip" "this" {
-  for_each = { for v in var.interfaces : v.name => v
-    if(!try(v.create_public_ip, false) && try(v.public_ip_name, null) != null)
+  for_each = { for v in var.interfaces : v.name => v if !v.create_public_ip && v.public_ip_name != null
   }
 
   name                = each.value.public_ip_name
-  resource_group_name = try(each.value.public_ip_resource_group, null) != null ? each.value.public_ip_resource_group : var.resource_group_name
+  resource_group_name = coalesce(each.value.public_ip_resource_group, var.resource_group_name)
```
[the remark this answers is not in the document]
FoSix: yes, good catch

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,243 @@
...
+  Firewall parameters configuration.
+
+  This map contains basic, as well as some optional Firewall parameters. Both types contain sane defaults.
+  Nevertheless they should be at least reviewed to meet deployment requirements.
+
+  List of either required or important properties: 
+
+  - `size`              - (`string`, optional, defaults to `Standard_D3_v2`) Azure VM size (type). Consult the *VM-Series
+                          Deployment Guide* as only a few selected sizes are supported
+  - `zone`              - (`number`, ??????????) Availability Zone to place the VM in, `null` value means a non-zonal deployment
```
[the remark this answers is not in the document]
FoSix: yeah, a left over after the moment where I did not know if we still have limitation on PIPs being only non-zonal or all-zones deployable

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,243 @@
...
+                          this Firewall will be created, explicit `null` means non-zonal deployment
+  - `disk_type`         - (`string`, optional, defaults to `StandardSSD_LRS`) type of Managed Disk which should be created,
+                          possible values are `Standard_LRS`, `StandardSSD_LRS` or `Premium_LRS` (works only for selected
+                          `vm_size` values)
+  - `disk_name`         - (`string`, optional, defaults to VM name + `-disk` suffix) name od the OS disk
```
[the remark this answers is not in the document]
FoSix: it's required in the resource, but in this variable definition it's optional, see line 139

modules/vmseries/variables.tf
```diff
@@ -1,219 +1,243 @@
...
+    username                        = optional(string, "panadmin")
+    password                        = optional(string)
+    disable_password_authentication = optional(bool, true)
+    ssh_keys                        = optional(list(string), [])
+  })
 }
 
-variable "avzones" {
+variable "image" {
   description = <<-EOF
```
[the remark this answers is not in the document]
FoSix: any specific place? or just in general?

modules/vmseries/main.tf
FoSix: supported in the current form

modules/vmseries/variables.tf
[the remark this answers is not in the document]
acelebanski: It seems fixed.

acelebanski (approved):
I'm fine with it now, let's fix the examples after we migrate to the new repo.

alperenkose (approved):
👍🏻

PR #372, opened 2023-12-20
refactor(module/gwlb): Refactor Gateway Load Balancer module

sebastianczech:
## Description

PR delivers changes for Gateway Load Balancer module.
New example with code similar to e.g. common VM-Series example, will be prepare in new PR.

## Motivation and Context

#328 

## How Has This Been Tested?

Code was tested by deploying `gwlb_with_vmseries` example, which is going to be delivered in new PR.

## Types of changes

## Checklist


modules/gwlb/main.tf
```diff
@@ -1,61 +1,64 @@
...
+  probe_threshold     = var.health_probe.probe_threshold
+  request_path        = var.health_probe.request_path
+  interval_in_seconds = var.health_probe.interval_in_seconds
+  number_of_probes    = var.health_probe.number_of_probes
 }
 
+# https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/lb_rule
 resource "azurerm_lb_rule" "this" {
-  name            = try(var.lb_rule.name, azurerm_lb.this.frontend_ip_configuration[0].name)
+  name            = coalesce(var.lb_rule.name, azurerm_lb.this.frontend_ip_configuration[0].name)
```
acelebanski, line 52: Ahh, now I see why this property was optional inside the variables file. Such name generation probably complies with our approach, perhaps to be discussed?

examples/gwlb_with_vmseries/variables.tf
```diff
@@ -264,7 +264,7 @@ variable "load_balancers" {
 
     Please refer to [module documentation](../../modules/loadbalancer/README.md#frontend_ips) for available properties.
 
-    > [!NOTE] 
+    > [!NOTE]
```
acelebanski, line 267: This admonition will not show properly, use this instead (like in `panorama` code for example):
`**Note!** \`

examples/gwlb_with_vmseries/example.tfvars
acelebanski, line 1: As discussed on Slack, entire example code to be discussed on call once you're back.

acelebanski (changes requested):
Left some comments on the module code, example to be discussed on call once you're back.

modules/gwlb/variables.tf
```diff
@@ -1,66 +1,143 @@
...
+    port                = number
+    protocol            = optional(string, "Tcp")
+    interval_in_seconds = optional(number)
+    probe_threshold     = optional(number)
+    request_path        = optional(string)
+    number_of_probes    = optional(number)
+  })
+  validation { # port
+    condition     = var.health_probe.protocol == "Tcp" ? var.health_probe.port != null : true
+    error_message = "The `port` property is required when protocol is set to \"Tcp\"."
```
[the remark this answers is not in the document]
sebastianczech: yes, right. 
Nevertheless I set `port` as required. Do you have proposition for better approach in that case?

modules/gwlb/variables.tf
```diff
@@ -1,66 +1,143 @@
...
+  Following settings are available:
+  - `name`                - (`string`, optional) name of the health probe. Defaults to `name` variable value.
+  - `port`                - (`number`, required) port to run the probe against
+  - `protocol`            - (`string`, optional, defaults to `Tcp`) protocol used by the health probe,
+                            can be one of "Tcp", "Http" or "Https".
+  - `probe_threshold`     - (`number`, optional) number of consecutive probes that decide on forwarding traffic to an endpoin.
+  - `request_path`        - (`string`, optional) used only for non `Tcp` probes,
+                            the URI used to check the endpoint status when `protocol` is set to `Http(s)`
+  - `interval_in_seconds` - (`number`, optional) interval in seconds between probes, with a minimal value of 5
+  - `number_of_probes`    - (`number`, optional)
```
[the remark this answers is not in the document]
sebastianczech: Added.

modules/gwlb/variables.tf
```diff
@@ -1,66 +1,143 @@
...
-  
-  For details, please refer to [provider documentation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/lb_probe#argument-reference).
+  Health probe configuration for the Gateway Load Balancer backends.
+
+  Following settings are available:
+  - `name`                - (`string`, optional) name of the health probe. Defaults to `name` variable value.
+  - `port`                - (`number`, required) port to run the probe against
+  - `protocol`            - (`string`, optional, defaults to `Tcp`) protocol used by the health probe,
+                            can be one of "Tcp", "Http" or "Https".
+  - `probe_threshold`     - (`number`, optional) number of consecutive probes that decide on forwarding traffic to an endpoin.
```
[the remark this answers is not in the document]
sebastianczech: Fixed.

modules/gwlb/variables.tf
```diff
@@ -1,66 +1,143 @@
...
+  }
+  nullable = false
+  type = object({
+    name                = optional(string)
+    port                = number
+    protocol            = optional(string, "Tcp")
+    interval_in_seconds = optional(number)
+    probe_threshold     = optional(number)
+    request_path        = optional(string)
+    number_of_probes    = optional(number)
```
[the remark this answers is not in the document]
sebastianczech: I missed that, but I fixed it.

modules/gwlb/variables.tf
```diff
@@ -1,66 +1,143 @@
...
-  - `name`              - (Optional|string) Name of the backend. If not specified name is generated from `name` variable and backend key.
-  - `tunnel_interfaces` - (Required|map) Map with tunnel interfaces specs.)
-
-  Each tunnel interface specification consists of following settings (refer to [provider documentation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/lb_backend_address_pool#tunnel_interface) for details):
-  - `identifier` - (Required|int) Interface identifier.
-  - `port`       - (Required|int) Interface port.
-  - `type`       - (Required|string) Either "External" or "Internal".
-  
+  - `name`              - (`string`, optional) name of the backend.
+                          If not specified name is generated from `name` variable and backend key.
```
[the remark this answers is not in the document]
sebastianczech: Right, fixed.

modules/gwlb/variables.tf
```diff
@@ -1,66 +1,143 @@
...
+  - `tunnel_interfaces` - (`map`, required) map with tunnel interfaces.
+
+  Each tunnel interface specification consists of following settings:
```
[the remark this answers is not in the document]
sebastianczech: I don't recall , that we agreed rules for nested maps, but nevertheless I don't see nested maps in `panorama` module - could you please put me URL to example of standard, which I should follow ?

modules/gwlb/variables.tf
```diff
@@ -1,66 +1,143 @@
...
   On GWLB inspection enabled VM-Series instance, `identifier` and `port` default to:
   - `800`/`2000` for `Internal` tunnel type
   - `801`/`2001` for `External` tunnel type
-  Variable default reflects this configuration on GWLB side. Additionally, for VM-Series tunnel interface protocol is always VXLAN.
+
+  Variable default reflects this configuration on GWLB side.
+  Additionally, for VM-Series tunnel interface protocol is always VXLAN.
   EOF
   default = {
     ext-int = {
```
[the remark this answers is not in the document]
sebastianczech: For me it is intuitive, but if you prefer different one, what name I should use?

modules/gwlb/variables.tf
```diff
@@ -80,21 +157,47 @@ variable "backends" {
...
-  description = "Azure tags to apply to the created resources."
+  Available options:
+  - `name`              - (`string`, optional) name for the rule.
+  - `load_distribution` - (`string`, optional, defaults to `Default`) specifies the load balancing distribution type
+                          to be used by the Gateway Load Balancer.
+  EOF
   default     = {}
-  type        = map(string)
+  nullable    = false
+  type = object({
```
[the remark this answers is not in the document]
sebastianczech: We cannot use `optional` for object - here is the error:

```
Invalid type specification: Keyword "optional" is valid only as a modifier for object type attributes.
```

Frankly speaking I've done it by purpose as we have always 1 rule , which is required , so I propose to leave it as it is - do you agree?

modules/gwlb/main.tf
```diff
@@ -1,61 +1,64 @@
...
-
-  tags = var.tags
 }
 
+# https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/lb_backend_address_pool
 resource "azurerm_lb_backend_address_pool" "this" {
   for_each = var.backends
 
-  name            = try(each.value.name, "${var.name}-${each.key}")
+  name            = coalesce(each.value.name, "${var.name}-${each.key}")
```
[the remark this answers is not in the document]
sebastianczech: Let's discuss this in next week (the same argument in 1 comment before).

modules/gwlb/main.tf
sebastianczech: Yes, let's discuss it.

examples/gwlb_with_vmseries/example.tfvars
sebastianczech: Yes, let's discuss it in next week.

examples/gwlb_with_vmseries/variables.tf
sebastianczech: Fixed.

modules/gwlb/variables.tf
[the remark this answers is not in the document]
acelebanski: While TCP is the default, I think that's the best approach.

modules/gwlb/variables.tf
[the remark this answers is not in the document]
acelebanski: Take a look at `autoscaling_profiles` variable in [vmss module](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/307-refactor-modules/modules/vmss/variables.tf).

modules/gwlb/variables.tf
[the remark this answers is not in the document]
acelebanski: You set both internal and external tunnel interfaces inside this map, that's why word `ext` concerns me. Perhaps we could change name `ext_int` to just `backend`?

modules/gwlb/variables.tf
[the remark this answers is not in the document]
acelebanski: So you are able to create a rule without a name, correct?

examples/gwlb_with_vmseries/.header.md
```diff
@@ -0,0 +1,62 @@
+# VM-Series Azure Gateway Load Balancer example
+
+The exmaple allows to deploy VM-Series firewalls for inbound and outbound traffic inspection utilizing Azure Gateway Load Balancer in service chain model as described in the following [document](https://docs.paloaltonetworks.com/vm-series/10-2/vm-series-deployment/set-up-the-vm-series-firewall-on-azure/deploy-the-vm-series-firewall-with-the-azure-gwlb).
```
FoSix, line 3: follow the 130 columns limit

examples/gwlb_with_vmseries/.header.md
```diff
@@ -0,0 +1,62 @@
...
+* Copy `example.tfvars` to `terraform.tfvars` and adjust it to your needs.
+* Copy `files/init-cfg.txt.sample` to `files/init-cfg.txt` and fill it in with required bootstrap parameters (see this [documentation](https://docs.paloaltonetworks.com/vm-series/10-2/vm-series-deployment/bootstrap-the-vm-series-firewall/create-the-init-cfgtxt-file/init-cfgtxt-file-components) for details).
+* (optional) Authenticate to AzureRM, switch to the Subscription of your choice if necessary.
+* Initialize the Terraform module:
+
+      terraform init
+
+* (optional) Plan you infrastructure to see what will be actually deployed:
+
+      terraform plan
```
FoSix, line 19: this notation is less error prone when doing conversions from MD to MDX/Docusaurus 

````suggestion
  ```bash
  terraform plan
  ```
````

File: examples/gwlb_with_vmseries/variables.tf [hunk 14]
```diff
@@ -112,10 +112,10 @@ variable "gateway_load_balancers" {
   - `name`                - (required|string) Gateway Load Balancer name.
   - `vnet_key`            - (required|string) Key of a VNet from `var.vnets` that contains target Subnet for LB's frontned. Used to get Subnet ID in combination with `subnet_key` below.
```
FoSix, line 113: 130 columns limit

examples/gwlb_with_vmseries/variables.tf
```diff
@@ -112,10 +112,10 @@ variable "gateway_load_balancers" {
...
   - `subnet_key`          - (required|string) Key of a Subnet from `var.vnets[vnet_key]`.
-  - `frontend_ip_config`  - (optional|map) Remaining Frontned IP configuration.
+  - `frontend_ip`         - (optional|map) Remaining Frontned IP configuration.
   - `resource_group_name` - (optional|string) LB's Resource Group, by default the one specified by `var.resource_group_name`.
   - `backends`            - (optional|map) LB's backend configurations.
-  - `heatlh_probe`        - (optional|map) Health probe configuration.
+  - `health_probes`       - (optional|map) Health probes configuration.
 
   Please consult [module documentation](../../modules/gwlb/README.md) for details.
   EOF
```
FoSix, line 121: missing type definition

examples/gwlb_with_vmseries/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 1.0.0, < 2.0"
+  required_version = ">= 1.3, < 2.0"
```
FoSix, line 2: wasn't the lower limit supposed to be `1.5` ?

modules/gwlb/.header.md
```diff
@@ -0,0 +1,79 @@
+# Gateway Load Balancer Module for Azure
+
+A Terraform module for deploying a Gateway Load Balancer for VM-Series firewalls.
+
+## Usage
+
+In order to use GWLB, below minimal definition of Gateway Load Balancer can be used, for which:
```
FoSix, line 7: Please follow the MD conventions, new line before a list

modules/gwlb/variables.tf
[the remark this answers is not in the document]
FoSix: can't we have this set up like in the regular [LB](https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/a63213f80532ba637c7d785f53d70c642a0ee59a/modules/loadbalancer/variables.tf#L338)?
- name - required, as in modules we do not autogenerate anything
- protocol - required, you take the decision what kind of protocol you want to use
- port - optional when HTTP(s) is used, then it defaults to protocol's port, required for TCP -> this logic can be handled in validation blocks

modules/gwlb/variables.tf
```diff
@@ -1,69 +1,148 @@
...
-  type        = map(any)
+  default = {
+    port = 80
+  }
+  nullable = false
+  type = object({
+    name                = optional(string)
+    port                = number
+    protocol            = optional(string, "Tcp")
+    interval_in_seconds = optional(number, 15)
```
FoSix, line 86: I would also keep this the same as in LB module -> use Azure defaults

FoSix: do not use your own defaults, unless Azure ones do not make sense in our deployments -> less maintenance

modules/gwlb/variables.tf
```diff
@@ -80,21 +159,47 @@ variable "backends" {
...
+  type = map(object({
+    name = string
+    tunnel_interfaces = map(object({
+      identifier = number
+      port       = number
+      protocol   = optional(string, "VXLAN")
+      type       = string
+    }))
+  }))
+  validation { # protocol
```
FoSix, line 172: if we can only have two backends, I would add a validation for that

modules/gwlb/variables.tf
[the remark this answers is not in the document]
FoSix: but the description is still wrong

modules/gwlb/variables.tf
```diff
@@ -1,69 +1,148 @@
...
 
   On GWLB inspection enabled VM-Series instance, `identifier` and `port` default to:
   - `800`/`2000` for `Internal` tunnel type
   - `801`/`2001` for `External` tunnel type
-  Variable default reflects this configuration on GWLB side. Additionally, for VM-Series tunnel interface protocol is always VXLAN.
+
+  Variable default reflects this configuration on GWLB side.
+  Additionally, for VM-Series tunnel interface protocol is always VXLAN.
   EOF
   default = {
```
FoSix, line 143: should we have a default value here? shouldn't it be moved to examples? In other words, is this default value specific to this module? or maybe to the examples?

modules/gwlb/variables.tf
```diff
@@ -80,21 +159,47 @@ variable "backends" {
...
-  type        = map(string)
-}
+  Load balancing rule configuration.
 
-variable "tags" {
-  description = "Azure tags to apply to the created resources."
+  Available options:
+  - `name`              - (`string`, optional) name for the rule.
+  - `load_distribution` - (`string`, optional, defaults to `Default`) specifies the load balancing distribution type
+                          to be used by the Gateway Load Balancer.
```
FoSix, line 193: I would add here a list of available options, or a link to external documentation (tf provider or azure) where that list can be found.

that's a general comment, for all variables add accept a finite list of arguments

modules/gwlb/variables.tf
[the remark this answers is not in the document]
FoSix: no, you cannot, according to provider's documentation, `name` [is required](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/lb_rule#name)

if we want to have this as optional i would suggest something like this

```hcl
default = null
nullable = true
type = object({
  name = string
  load_distribution = optional(string, "Default")
})
```

This would require minor changes in the code to handle `null`

modules/gwlb/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 1.0, < 2.0"
+  required_version = ">= 1.3, < 2.0"
```
FoSix, line 2: please bump tf trequirements to min = 1.5 and provider to min 3.80

FoSix (changes requested):

modules/gwlb/variables.tf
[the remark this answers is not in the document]
sebastianczech: I removed `If not specified name is generated from `name` variable and backend key.`, thanks

modules/gwlb/variables.tf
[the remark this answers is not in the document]
sebastianczech: Thanks, I adjusted nested maps indentation .

modules/gwlb/variables.tf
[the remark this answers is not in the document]
sebastianczech: Ok, I replaced `ext-int` by `backend`.

modules/gwlb/variables.tf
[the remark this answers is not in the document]
sebastianczech: no, I was not creating a rule without name - in case name was not provided, I used:

`name = coalesce(var.lb_rule.name, azurerm_lb.this.frontend_ip_configuration[0].name)`

Nevertheless I'm going to change as required to have the same approach for naming as for other resources.

modules/gwlb/main.tf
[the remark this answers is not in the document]
sebastianczech: I've made `name` as required, name generation logic is moved to the example.

modules/gwlb/main.tf
sebastianczech: I've made `name` as required, name generation logic is moved to the example.

examples/gwlb_with_vmseries/.header.md
sebastianczech: Fixed

examples/gwlb_with_vmseries/.header.md
sebastianczech: Fixed

modules/gwlb/.header.md
sebastianczech: Fixed

modules/gwlb/variables.tf
sebastianczech: I removed default value for `protocol`, but I don't think that for LB we are following Azure defaults e.g.
- [interval_in_seconds](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/lb_probe#interval_in_seconds) have default value 15, but in LB it's missed https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/ad1c4d7ff2cad59a59cb1fe3bf820e65a2586b88/modules/loadbalancer/variables.tf#L357
- [protocol](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/lb_probe#protocol) is optional, not required as in https://github.com/PaloAltoNetworks/terraform-azurerm-vmseries-modules/blob/ad1c4d7ff2cad59a59cb1fe3bf820e65a2586b88/modules/loadbalancer/variables.tf#L354

modules/gwlb/variables.tf
sebastianczech: Added

modules/gwlb/variables.tf
sebastianczech: I added default value for variable `backends` to have it as it's described in https://docs.paloaltonetworks.com/vm-series/10-2/vm-series-deployment/set-up-the-vm-series-firewall-on-azure/deploy-the-vm-series-firewall-with-the-azure-gwlb, but if you don't agree with that approach, I can remove it - can we discuss it ?

modules/gwlb/versions.tf
sebastianczech: done

modules/gwlb/variables.tf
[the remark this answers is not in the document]
sebastianczech: Done

modules/gwlb/variables.tf
sebastianczech: Done

examples/gwlb_with_vmseries/example.tfvars
sebastianczech: Example is going to be delivered in new PR.

examples/gwlb_with_vmseries/variables.tf
sebastianczech: Example is going to be delivered in new PR.

